PR #8, opened 2019-01-24
Tiny memory leak in tag families implementation

okazmin:
There is code 
`tf->name = strdup("tag25h9");`
in each tag family implementation. According to specification of `strdup` function (https://en.cppreference.com/w/c/experimental/dynamic/strdup) the pointer returned by it should be freed by user of the function. But tag_family_destroy does not do it. It leads to tiny memory leak (about 9 byte each) but still.
I assume that those changes should be done also in generator of families implementation.


mkrogius:
Thanks for the PR!

PR #11, opened 2019-02-24
CMake support

christianrauch:
This PR adds support for the CMake build system and pkg-config. This enables the library to be used in other CMake projects. I also added a `package.xml` with metadata about the library. It is commonly used by other build systems (e.g. catkin and colcon for ROS1 and ROS2) to track build dependencies.

I used to maintain this in a separate git repository at https://github.com/christianrauch/apriltag2 before the AprilTags library moved to a versioning system. I would like to have this merged upstream to ease updates of the library and the use in other build systems.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -1,0 +1,21 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="2">
+  <name>apriltag</name>
+  <version>0.10.0</version>
+  <description>AprilTag detector library</description>
+
+  <maintainer email="mkrogius@umich.edu">Max Krogius</maintainer>
+
+  <author email="ebolson@umich.edu">Edwin Olson</author>
+  <author email="mkrogius@umich.edu">Max Krogius</author>
+
+  <license>BSD</license>
+  <url>https://april.eecs.umich.edu/software/apriltag.html</url>
+
+  <buildtool_depend>cmake</buildtool_depend>
+
+  <export>
+    <build_type>cmake</build_type>
+  </export>
+</package>
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
+*.a
+*.so
 *.o
 *.swp
 example/opencv_demo
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,44 @@
+cmake_minimum_required(VERSION 3.1)
+project(apriltag)
+
+find_package(Threads REQUIRED)
+find_package(OpenCV QUIET)
+
+include_directories(.)
+aux_source_directory(common COMMON_SRC)
+aux_source_directory(. APRILTAG_SRCS)
+
+# Library
+add_library(apriltag SHARED ${APRILTAG_SRCS} ${COMMON_SRC})
+target_link_libraries(apriltag Threads::Threads m)
+file(GLOB_RECURSE HEADER_FILES RELATIVE ${CMAKE_SOURCE_DIR} *.h)
+
+# Examples
+# apriltag_demo
+add_executable(apriltag_demo example/apriltag_demo.c)
+target_link_libraries(apriltag_demo apriltag)
+
+# opencv_demo
+if(OpenCV_FOUND)
+    add_executable(opencv_demo example/opencv_demo.cc)
+    target_link_libraries(opencv_demo apriltag ${OpenCV_LIBRARIES})
+    install(TARGETS opencv_demo RUNTIME DESTINATION bin)
+endif(OpenCV_FOUND)
+
+# install library and example programs
+install(TARGETS apriltag apriltag_demo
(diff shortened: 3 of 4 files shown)


mkrogius:
Hi,

Thanks for maintaining that repo, and sorry for the delay in responding to this pull request.

I guess I'm not sure that I am convinced that we necessarily need cmake for this project. In most cases I would imagine the easiest thing to do for a project that uses apriltag would be to install apriltag separately rather than building it as part of said projects build. What do you see as the advantages of switching to Cmake?. The nice thing about make is that installation is simple for our users. 

Thanks,
Max

christianrauch:
I would say that the biggest advantage of CMake is the platform-independent generation of make or project files. E.g. it automatically determines the compiler and the target library file format (*.so, *.dll, ...). These files are usually also generated out of tree, which helps to keep the source directory free of cluttered temporary files.

However, I actually do not want to convince you to replace your Makefile with this CMake file. I rather want to add CMake support (alongside the Makefile) to allow build systems that use CMake (e.g. other cmake based projects, catkin, colcon) to directly use the upstream project.
This PR provides additional build system support and does not interfere with the Makefile.

wxmerkt:
I think adding a CMake buildfile and package.xml to the library is a big plus as it's a modern build system that easily integrates with other libraries and correctly configures/sets up the Makefile for each platform. Maybe it's worth to add an additional `apriltags-config.cmake` generation to the above file so that `find_package(apriltag)` can find it?

allsey87:
Just to add my two cents. I have used CMake and Yocto to create a libapriltags that I deploy on my embedded target. This is very useful since Yocto takes care of all the complications of cross compilation. If anyone is interested, see the [Bitbake recipe](https://github.com/allsey87/meta-builderbot/blob/master/recipes-robot/libapriltag/libapriltag_2016.12.01.bb) and the [CMakeLists.txt](https://github.com/allsey87/meta-builderbot/blob/master/recipes-robot/libapriltag/files/CMakeLists.txt) file.

mkrogius:
Ok, it seems like we've got plenty of people who want Cmake so I'll accept this PR. Unfortunately I can't merge this just yet since we are currently in the process of figuring out copyright legalese, but we should get that figured out quite soon.

package.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="2">
+  <name>apriltag</name>
```
paulbovbel: Should this be `apriltags` to be consistent with the repository name?

paulbovbel:
@wxmerkt I guess this eliminates the need for https://github.com/dmalyuta/apriltags2_ros/pull/43 and submodules/externalproject, since with cmake and package.xml, apriltags will be trivial to release into ROS as deb packages (whoo!).

Who is going to own releasing apriltags into ROS? That person should probably be listed as the/a maintainer in `package.xml`.

allsey87:
A small request on this PR, when merging in the CMake build script, please consider setting the following variable correctly. It is needed by the Yocto build system:

```cmake
set_target_properties(apriltag PROPERTIES SOVERSION X.X.X VERSION Y.Y.Y)
```

wxmerkt:
I am happy to do the release if nobody else volunteers (and the release repository question and access is settled). Once this is done I will incorporate it in https://github.com/dmalyuta/apriltags2_ros/pull/43.

package.xml
christianrauch: I named it `apriltag` according to the ICRA 2011 paper and the description on the webside:

> AprilTag is a visual fiducial system ...

What is the official name? I.e. I haven't seen the plural `apriltags` as a proper name in any of the references.

christianrauch:
I released the apriltag library and ROS2 node to crystal: https://github.com/ros/rosdistro/blob/73477b86d9bef59afeea8e368a835b4b83f10fcd/crystal/distribution.yaml#L148-L182. I would prefer if the release repository for ROS1 and ROS2 could be hosted in an official place (e.g. in `AprilRobotics` or `ros-gbp`). Since this is a plain cmake package, there should be no issue releasing it to both from the same branch.

@allsey87 Feel free to send request directly as PR to my branch.

package.xml
paulbovbel: Right, just looking at the repository name, which conventionally matches up with the ROS package name.

paulbovbel:
Sounds good, I can organize a repo in ros-gbp once this is merged.

allsey87:
@christianrauch should the PR be made against this branch? https://github.com/christianrauch/apriltag2

christianrauch:
@allsey87 Can you make your PR against branch `cmake` of my fork of the `apriltags` repo: https://github.com/christianrauch/apriltags/tree/cmake (i.e. the branch and fork this very PR is based on).

mkrogius:
I have merged this PR in. Perhaps lets create new issues on this repo to track the different pieces of follow on work that people are talking about here.

PR #12, opened 2019-03-07
adding 36h9

davelkan:
adding 36h9 tag library


mkrogius:
Hi David, I'm not sure we want to add extra families to the distribution of AprilTags, especially since as far as I can tell this library was not included with previous releases. Why do you think that this family should be added?

Best,
Max

davelkan:
I primarily use this library so I've been using forked versions to incorporate it. For me it would be convenient to include it, and I don't see much cost to doing so.

mkrogius:
This library is the main AprilTags library so changes to this library must be justified in terms of the benefit to the overall community. Adding a tag family does increase the size of the library, so if it is not uniquely useful compared to the other tag families then I will not approve its addition.

If it is a pain to maintain a fork, I would suggest that you instead install AprilTags normally and keep this file in your own source code instead of building it into the AprilTags library. Then it should be painless to incorporate upstream changes!

By the way it looks like this file will not be compatible with AprilTag3. If you have the original java class for this file then you should generate a new .c file following the instructions here: https://github.com/AprilRobotics/apriltag-generation

PR #17, opened 2019-03-22
Fix memory leaks

s-trinh:
This should fix some detected memory leaks with valgrind.


lecramdev:
There is another leak with debug set to one. The darker image is not destroyed when generating the "debug_output.pnm" (apriltag.c, l.1275-1310).

s-trinh:
@lecramdev 

Thanks, I have added the fixes for the memory leaks detected with debug option.

mkrogius:
Thanks for the fixes!

PR #20, opened 2019-03-31
export project

christianrauch:
This PR exports the `apriltag` library for other cmake projects.

One can use the library in other cmake projects by adding these two lines:
```CMake
find_package(apriltag REQUIRED)
target_link_libraries(<your_project> apriltag)
```

The include and library directories are automatically added by the exported properties.


mkrogius:
Thanks for the PR!

PR #23, opened 2019-04-06
Check for NULL before calling matd_destroy

### LINKED ISSUE #22: Error, more than one new minima found.
Hello, 
I edited the opencv_demo by adding the following because I am interested in pose:

//POSE DETECTION:
apriltag_detection_info_t info;
info.det = det;
info.tagsize = 100;
info.fx = 630;
info.fy = 630;
info.cx = 640;
info.cy = 360;
apriltag_pose_t pose;
double err = estimate_tag_pose(&info, &pose);
cout << "Err: " << err <<endl;

When I execute opencv_demo, it works for about 5 seconds, then crashes with this error:

Error, more than one new minima found.
*** Error in `./opencv_demo': free(): invalid pointer: 0x00007f01cd5ad81f ***
======= Backtrace: =========

Then it lists a bunch of locations.
How can I fix this?

s-trinh:
`pose2.R` is initialized only if there is one new minima:
https://github.com/AprilRobotics/apriltags/blob/a5ab81dd79fd294b77940efad889a820509e83ae/apriltag_pose.c#L412-L426

`pose2.t` is initialized only if `pose2.R` is not `NULL`:
https://github.com/AprilRobotics/apriltags/blob/a5ab81dd79fd294b77940efad889a820509e83ae/apriltag_pose.c#L496-L502

Probably should fix #22

### DIFF
--- a/apriltag_pose.c
+++ b/apriltag_pose.c
@@ -517,8 +517,10 @@ double estimate_tag_pose(apriltag_detection_info_t* info, apriltag_pose_t* pose)
     if (err1 <= err2) {
         pose->R = pose1.R;
         pose->t = pose1.t;
+        if (pose2.R) {
+            matd_destroy(pose2.t);
+        }
         matd_destroy(pose2.R);
-        matd_destroy(pose2.t);
         return err1;
     } else {
         pose->R = pose2.R;


mkrogius:
Thanks, good catch!

PR #27, opened 2019-04-17
Support building on Windows

josch:
- with MSVC and vcpkg
 - with MingW
 - with Cygwin

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,7 +1,6 @@
 cmake_minimum_required(VERSION 3.1)
 project(apriltag)
 
-find_package(Threads REQUIRED)
 find_package(OpenCV QUIET)
 
 include_directories(.)
@@ -10,7 +9,15 @@ aux_source_directory(. APRILTAG_SRCS)
 
 # Library
 add_library(${PROJECT_NAME} SHARED ${APRILTAG_SRCS} ${COMMON_SRC})
-target_link_libraries(${PROJECT_NAME} Threads::Threads m)
+if (MSVC)
+    # FindThreads will not find pthread.h with MSVC
+    # winmm is necessary for __imp_timeGetTime
+    find_library(PTHREAD_LIBRARIES NAMES pthreads)
+    target_link_libraries(${PROJECT_NAME} ${PTHREAD_LIBRARIES} winmm)
+else()
+    find_package(Threads REQUIRED)
+    target_link_libraries(${PROJECT_NAME} Threads::Threads m)
+endif()
 
 set_target_properties(${PROJECT_NAME} PROPERTIES SOVERSION 3 VERSION 3.0.0)
 
--- a/apriltag.c
+++ b/apriltag.c
@@ -38,7 +38,6 @@ either expressed or implied, of the Regents of The University of Michigan.
 #include <string.h>
 #include <stdio.h>
 #include <inttypes.h>
-#include <sys/time.h>
 
 #include "common/image_u8.h"
 #include "common/image_u8x3.h"
@@ -57,6 +56,18 @@ either expressed or implied, of the Regents of The University of Michigan.
 
 #ifndef M_PI
 # define M_PI 3.141592653589793238462643383279502884196
+#endif
+
+#ifdef _WIN32
+static inline void srandom(unsigned int seed)
+{
+        srand(seed);
+}
+
+static inline long int random(void)
+{
+        return rand();
+}
 #endif
 
 #define APRILTAG_U64_ONE ((uint64_t) 1)
@@ -512,7 +523,7 @@ double value_for_pixel(image_u8_t *im, double px, double py) {
 }
 
 void sharpen(apriltag_detector_t* td, double* values, int size) {
-    double sharpened[size*size];
+    double *sharpened = malloc(sizeof(double)*size*size);
     double kernel[9] = {
         0, -1, 0,
         -1, 4, -1,
@@ -539,6 +550,8 @@ void sharpen(apriltag_detector_t* td, double* values, int size) {
             values[y*size + x] = values[y*size + x] + td->decode_sharpening*sharpened[y*size + x];
         }
     }
+
+    free(sharpened);
 }
 
 // returns the decision margin. Return < 0 if the detection should be rejected.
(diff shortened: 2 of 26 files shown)


mkrogius:
Hi @josch, it looks like a lot of this commit is removing all VLAs. Was this part of the change necessary for all the compilers or just for MSVC?

Also I see that you added definitions for srandom and random. Where are those functions being called from?

Best,
Max

josch:
Yes, sadly, MSVC does not support VLAs because it only conforms to ANSI C 9899:1990:

https://docs.microsoft.com/en-us/cpp/c-language/ansi-conformance?view=vs-2019
https://stackoverflow.com/questions/5246900/enabling-vlas-variable-length-arrays-in-ms-visual-c

I added definitions for srandom and random to those compilation units that uses them and only to those. That's also why I declared them as static. I thought this to be less intrusive than a common definition in another header because I didn't see a header that was used by all C files that use random or srandom. The problem here is, that windows.h does not have either random or srandom but only rand and srand. That's why this change applies to MSVC and MinGW (but not cygwin).

mkrogius:
I don't think we want to restrict AprilTag to C90 just for the sake of MSVC. It seems to me that if we support MinGW and Cygwin that that should be sufficient for building and using our library on windows. If someone wants to use AprilTags with a Visual Studio project they can build the dll using either of those compilers and then include it.

I will happily approve the MinGW and Cygwin related changes.

josch:
That would be unfortunate for us. Please reconsider. We had and have actual customers who require building the whole source with visual studio.

Which problems do you see by dropping VLAs in favour of malloc? Some trivial benchmarks I carried out did not see any performance difference.

I agree with you that it is very silly having to drop VLAs just to support the MS compiler. But please don't reject MSVC support out of personal feelings (which I share) but provide some technical reasons.

Thanks!

mkrogius:
I don't have too much against removing VLAs. Rather the concern here is about maintenance of the support for MSVC. I don't want to have to test future releases on windows or even think at all about whether the code I am writing is C11 vs C90.

josch:
Thank you for your understanding. How about the following: I remove the "support building with MSVC" from the commit message but leave in the VLA->malloc replacement. That way, support for MSVC is just "accidental" and might break at any point in the future. I totally agree that you shouldn't care whether the code you are writing is C11 or C90 but if you in principal have no problem with removing the current use of VLAs in favour of malloc then this would allow us to use apriltags unpatched for now. As far as I'm concerned it would be totally okay if apriltags breaks on windows in a future release. In that case I would just file another pull request. Just don't advertise that apriltags is supposed to build on windows and then you don't have to maintain it, right? Would that course of action be okay for you?

mkrogius:
Alright, that sounds good to me!

josch:
Wonderful! :)

I changed the commit message accordingly. Tell me if you'd like me to change anything else about this commit.

Thanks!

mkrogius:
Looks good, just one last change. Can you remove the cast of malloc/calloc in all the VLA-removal statements? I believe it is better practice to not cast malloc/calloc.

josch:
Also done!

Would you also appreciate a pull request adding a test using ctest? Running `apriltag_demo` against a pre-generated image helped me making sure that my changes didn't break anything.

mkrogius:
Sorry for the slow response, yes a ctest PR would be greatly appreciated!

catree:
@josch 

I am trying to build AprilTag3 with MSVC. Can you share how you do it?
What pthreads for Windows version do you use? Could you share a CMakeLists to build pthreads for Visual Studio?

Thanks.

josch:
@catree we use pthreads from vcpkg

catree:
@josch 

Thanks for the reference.

pfaion:
@josch 

I'm also trying to build AprilTag3 with MSVC. I installed pthreads from vcpkg, but it's not recognized from cmake. Did you modify the CMakeLists.txt? Any tips on how to make it work would be greatly appreciated!

Thanks!

josch:
Nope, no changes to AprilTag's CMakeLists.txt. This is our embedded copy: https://github.com/3DTK/3DTK/blob/master/3rdparty/apriltag-3.0/CMakeLists.txt

And this is the prove that it works just fine on MSVC, cygwin and mingw: https://ci.appveyor.com/project/josch/3dtk

This is the build script that we use with msvc: https://github.com/3DTK/3DTK/blob/master/windows/build.cmd As you can see, we do nothing special with pthreads.

pfaion:
Ok, thank you so much for your detailed response!

In case anyone else comes across this issue, I figured out the reason for pthreads from vcpkg not working:

I assume **josch** is not using the newest version of vcpkg. The CMakeLists.txt script in the apriltag library looks for the path to pthreads in the variable PTHREADS_LIBRARIES. This variable was defined in older versions of the vcpkg pthreads-port, but got dropped 2 months ago, which leads to apriltags CMake not finding the reference. Currently the variable is already added back into vcpkg, as you can see here: https://github.com/microsoft/vcpkg/blob/master/ports/pthreads/vcpkg-cmake-wrapper.cmake
BUT this fix got added shortly after the last release, so the current version of vcpkg does not yet include this fix.

I solved this by building pthreads4win manually and modifying the CMakeLists.txt from apriltags, but I assume the issue should be fixed with the next vcpkg release or you can just build the current version of vcpkg from source.

josch:
Correct, I was not using the the git HEAD of vcpkg but the latest released version installed by appveyor.

And there was indeed a change in the pthread library in vcpkg but even with the git HEAD of vcpkg I did not find a way to build with windows pthreads such that it would work with the old and the new version.

Right now we have apriltags patched such that it links against pthreads on windows like this:

```
if (MSVC)
    find_package(pthreads REQUIRED)
    target_include_directories(${PROJECT_NAME} PUBLIC ${PThreads4W_INCLUDE_DIR})
    target_link_libraries(${PROJECT_NAME} PThreads4W::PThreads4W winmm)
else()
    find_package(Threads REQUIRED)
    target_link_libraries(${PROJECT_NAME} Threads::Threads m)
endif()
```

AndreV84:
is there an easy to implement solution for Windows as for now?

PR #34, opened 2019-05-07
Added a cast from int to unsigned int in common/matd.h

avsegal:
This prevents compilation errors when including apriltag headers into external libraries which have -Werror and -Wall enabled.


mkrogius:
Thanks!

PR #39, opened 2019-06-10
Fix warnings detected with -Wsizeof-pointer-memaccess and -Wshadow

catree:
Example:

```
warning: declaration of ‘entry_idx’ shadows a previous local [-Wshadow]
         for (int entry_idx = 0; entry_idx < zh->nentries; entry_idx++) {
                  ^

```
and

```
warning: argument to ‘sizeof’ in ‘memset’ call is the same expression as the destination; did you mean to dereference it? [-Wsizeof-pointer-memaccess]
                             memset(dcpred, 0, sizeof(dcpred));
                                                     ^

```


mkrogius:
Thanks for the PR!

PR #47, opened 2019-07-18
Please have a look at the AprilTags.jl library 

Affie:
We have been maintaining it for some while now.

https://github.com/JuliaRobotics/AprilTags.jl


mkrogius:
Thanks for the PR and thanks for writing a Julia wrapper for AprilTags! Glad to see it!

PR #51, opened 2019-07-26
Remove unused argument to zarray_copy_data

macmason:
Because this function is implemented in the header, it leaks the need for `-Wno-unused-parameter` to any translation unit that includes it. As it's genuinely unused (in fact, the entire function is unused), may as well take it out.


mkrogius:
Thanks! Merged.

PR #64, opened 2019-11-13
Add support for using Python 2

jonazpiazu:
In many scenarios (e.g. using ROS Kinetic) it might be convenient to use Python 2 instead of 3. This PR adds support for Python 2 by specifying ` -DPREFERRED_PYTHON_VERSION=2`.


mkrogius:
I would like to apologize for not making it around to this PR sooner, and thank you for taking the time to submit it. My free time to work on AprilTag is quite limited.

That said, I do not want to accept this PR because Python 2 is now no longer supported: https://www.python.org/doc/sunset-python-2/

Best,
Max

PR #65, opened 2019-12-15
use FILES rules to install python wrapper

christianrauch:
If apriltag is installed locally and `~/.local/lib/python3.6` does not exist, the installation will fail with: `cp: cannot create regular file '/home/ubuntu/.local/lib/python3.6/site-packages': No such file or directory`. 

This is caused by
https://github.com/AprilRobotics/apriltag/blob/cad1009f9fbf457e2a5393c0fb6c8a05042a0611/CMakeLists.txt#L96
since files are simply copied via the `cp` command, which does not automatically create the file structure.

This PR installs this file via https://cmake.org/cmake/help/v3.13/command/install.html#installing-files which creates the necessary file hierarchy.


mkrogius:
Thanks for the PR!

PR #86, opened 2020-06-12
Adding a static version of the AprilTag library

roxlu:
Minimal update which adds a static version of the AprilTag library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,6 +13,8 @@ set(CMAKE_BUILD_TYPE Release)
 file(GLOB TAG_FILES ${CMAKE_SOURCE_DIR}/tag*.c)
 set_source_files_properties(SOURCE ${TAG_FILES} PROPERTIES COMPILE_FLAGS -O0)
 add_library(${PROJECT_NAME} SHARED ${APRILTAG_SRCS} ${COMMON_SRC} ${TAG_FILES})
+add_library(${PROJECT_NAME}_static STATIC ${APRILTAG_SRCS} ${COMMON_SRC} ${TAG_FILES})
+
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -13,6 +13,8 @@ set(CMAKE_BUILD_TYPE Release)
 file(GLOB TAG_FILES ${CMAKE_SOURCE_DIR}/tag*.c)
 set_source_files_properties(SOURCE ${TAG_FILES} PROPERTIES COMPILE_FLAGS -O0)
 add_library(${PROJECT_NAME} SHARED ${APRILTAG_SRCS} ${COMMON_SRC} ${TAG_FILES})
+add_library(${PROJECT_NAME}_static STATIC ${APRILTAG_SRCS} ${COMMON_SRC} ${TAG_FILES})
```
simogasp, line 16: it would be better to avoid this and remove `SHARED` from the orginal code and let the standard variable `BUILD_SHARED_LIBS` to drive the build for shared or static version

christian-rauch: This is already proposed in https://github.com/AprilRobotics/apriltag/pull/172.

christian-rauch:
An alternative to this is proposed in https://github.com/AprilRobotics/apriltag/pull/172. @roxlu Can you check if this PR also works for you by setting `BUILD_SHARED_LIBS` to `OFF`?

CMakeLists.txt
christian-rauch: The proposed change has been merged. You can use `BUILD_SHARED_LIBS` now to control the type of library.

christian-rauch:
Superseded by #172.

Issue #87, opened 2020-06-12
release to noetic and foxy


christian-rauch:
@wxmerkt Could we release the library to `noetic` (ROS1) and `foxy` (ROS2)?

Since `noetic` will be the last ROS1 version ever and `foxy` is supported for 3 years, I am not going to ask for further releases for a while after this :-)

wxmerkt:
Absolutely, happy to make the release.

@mkrogius, do we want to tag a new release since there have been 29 commits to master since the last version tag 3.1.1? This also requires an update of the version number in [package.xml](https://github.com/AprilRobotics/apriltag/blob/cad1009f9fbf457e2a5393c0fb6c8a05042a0611/package.xml#L5).

mkrogius:
Done, bumped to 3.1.2

wxmerkt:
Thank you for the swift action! I noticed the changes are more than bugfixes as also a Python3 wrapper was added. 

I've made the releases and will monitor what comes back from the buildfarm. I feel we may have to do patches or fixes for GNUInstallDirs or due to the addition of the Python wrapper.

Here are the rosdistro PRs:
- Noetic: https://github.com/ros/rosdistro/pull/25455
- Foxy: https://github.com/ros/rosdistro/pull/25456

Please note these releases of 3.1.2 do not include the Python wrapper.

https://github.com/AprilRobotics/apriltag/pull/88 fixes the dependency on NumPy which will then in turn build the Python wrappers in the next release.

wxmerkt:
Builds on both Noetic and Foxy are now passing - we can close this issue.

christian-rauch:
Thanks @wxmerkt !

PR #88, opened 2020-06-13
Fix dependencies, update CI

wxmerkt:
Relates to #87

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,11 +3,14 @@
 sudo: required
 language: generic
 env:
+  global:
+    - CATKIN_SKIP_TESTING=1
+    - PRERELEASE=1
   matrix:
-    - ROS_DISTRO="kinetic" PRERELEASE=true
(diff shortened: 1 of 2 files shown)


wxmerkt:
Before we merge this: I am wondering if we want to enforce a Python3 wrapper for ROS1 distros prior to Noetic at all. It won't work proper (as ROS1<noetic == Python2) and will complicate the release process. It will also pull in the python, python-numpy dependencies by default (i.e., greater install size for an otherwise small library). As thus I am wondering whether to deactivate the `build_depend` keys and just default to not building the wrapper when those are not available (behaviour < 3.1.2). Thoughts?

mkrogius:
Hi, sorry for not responding sooner. I am definitely on board with not building the python wrapper by default in these cases.

wxmerkt:
Thanks for the feedback - this is good to merge now. :-)

mkrogius:
Great, thanks for putting this together!

PR #89, opened 2020-06-17
add appveyor.yml

josch:
Build and test apriltags on windows with visual studio and publish artifacts.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,12 +12,21 @@ set(CMAKE_BUILD_TYPE Release)
 # Library
 file(GLOB TAG_FILES ${CMAKE_SOURCE_DIR}/tag*.c)
 set_source_files_properties(SOURCE ${TAG_FILES} PROPERTIES COMPILE_FLAGS -O0)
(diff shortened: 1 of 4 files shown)


christian-rauch:
Thanks for this contribution. I think it's a nice addition to be able to build the `apriltag` library on Windows and have another compiler to check for code issues. Thanks for the contribution.

The repo is already using GitHub Actions. It would be better to just extend the [GitHub Actions configuration](https://github.com/AprilRobotics/apriltag/blob/master/.github/workflows/industrial_ci_action.yml) to also build for Windows. You can use my [ROS Qt Creator plugin configuration](https://github.com/ros-industrial/ros_qtc_plugin/blob/devel/.github/workflows/main.yml) for inspiration to build a project on multiple platforms.

The main contribution is, from my point of view, the build on Windows. Your PR and the commit does not reflect this properly. Can you split the PR into two commits, one for enabling the WIndows build and one for adding the CI configuration?

christian-rauch:
I noticed that this included @josch's other PR https://github.com/AprilRobotics/apriltag/pull/27, which already adds Windows support. This PR is now just about adding another CI configuration. Since we are already using GitHub Actions, it would be preferable to add the Windows CI there.

PR #95, opened 2020-06-29
Apriltag library build in Windows without pthreads or winmm (option to remove OpenCV and Python added in CMake)

ryancwicks:
I've updated the cmake files and added some wrappers around the threading to allow the library to build without pthreads. I've also updated the CMake file so the library can be build without examples and python bindings.

Changes include:
-Setting the examples and python builds as option in CMake (default is on to maintain same behaviour)
-Wrapped windows threading in pthreads  to allow for dependency free builds in Windows. Based on blog post: https://nachtimwald.com/2019/04/05/cross-platform-thread-wrapper/
-Removed dependency on winmm.dll by removing timeGetTime call in time_util.c
-It now continues to build in Linux.


ooeygui:
Thank you!
I was looking to add apriltag to vcpkg for use on a Hololens 2. pthreads are not supported in UWP, so this change is really appreciated!

ooeygui:
I'm curious why the target was renamed to apriltag-umich?

ryancwicks:
It's a mistake on my part. I didn't realize that changes to my forked copy after submitting the pull request would continue to update. I'll fix it tonight.

christian-rauch:
@ryancwicks Can you rebase this and also add a CI workflow for Windows?

ryancwicks-voyis:
I'll take a pass at it this weekend.

.github/workflows/windows_ci_action.yml
```diff
@@ -0,0 +1,15 @@
...
+on: [push, pull_request] # on all pushes and PRs
+
+jobs:
+  CI_windows:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: recursive
+      - run: mkdir build; cd build; cmake ..; cmake --build . --config Release
```
christian-rauch, line 15: You can simplify this folder creation with: `cmake -B build -DCMAKE_BUILD_TYPE=Release && cmake --build build`

common/pjpeg.c
```diff
@@ -833,6 +833,12 @@ pjpeg_t *pjpeg_create_from_file(const char *path, uint32_t flags, int *error)
     uint8_t *buf = malloc(buflen);
     fseek(f, 0, SEEK_SET);
     int res = fread(buf, 1, buflen, f);
+
+    int error_code = ferror(f);
+    if (error_code ){
+        printf ("Read failed with error code: %d\n", error_code );
+    }
```
christian-rauch, line 840: I don't think the specific `error_code` is significant here. The documentation only says it's nonzero if there is an error. So this could be:
```cpp
if (ferror(f))
    debug_print("Read failed");
```
Use the new `debug_print` to report errors so they can be silenced in release builds. You can set the `errno` as proposed in https://github.com/AprilRobotics/apriltag/pull/224.

You may also have to use `clearerr` to clear the error.

common/pthreads_cross.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+typedef struct {
+    SRWLOCK lock;
+    bool    exclusive;
+} pthread_rwlock_t;
+
+//struct timespec {
+//    long tv_sec;
+//    long tv_nsec;
+//};
```
christian-rauch, line 14: Remove commented code.

common/pthreads_cross.cpp
```diff
@@ -0,0 +1,245 @@
...
+    if (abstime == NULL)
+        return INFINITE;
+
+    t = ((abstime->tv_sec - time(NULL)) * 1000) + (abstime->tv_nsec / 1000000);
+    if (t < 0)
+        t = 1;
+    return t;
+}
+
+#endif
```
christian-rauch, line 245: All files must have a new line at the end of the file.

common/pthreads_cross.cpp
```diff
@@ -0,0 +1,245 @@
...
+        ReleaseSRWLockShared(&(rwlock->lock));
+    }
+    return 0;
+}
+
+int sched_yield() {
+    return (int)SwitchToThread();
+}
+
+#ifdef _WIN32
```
christian-rauch, line 206: Maybe I am overlooking something here, but I think there is already a redundant `#ifdef _WIN32` before this.

common/workerpool.c
```diff
@@ -116,9 +116,8 @@ workerpool_t *workerpool_create(int nthreads)
         for (int i = 0; i < nthreads; i++) {
             int res = pthread_create(&wp->threads[i], NULL, worker_thread, wp);
             if (res != 0) {
-                debug_print("Insufficient system resources to create workerpool threads\n");
-                // errno already set to EAGAIN by pthread_create() failure
-                return NULL;
+                perror("pthread_create");
+                exit(-1);
```
christian-rauch, line 120: Don't print in the library in release mode. We added `debug_print` so that release builds will prevent stdout and stderr output. You also change the behaviour here by quitting the process instead of returning NULL. This was added in https://github.com/AprilRobotics/apriltag/pull/224 by design.

example/apriltag_demo.c
```diff
@@ -54,6 +60,10 @@ either expressed or implied, of the Regents of The University of Michigan.
 //
 // tagtest [options] input.pnm
 
+#ifdef _WIN32
+#define  hamm_hist_max 10
+#endif
+
```
christian-rauch, line 66: Why is this not a variable like for the Linux case? Use the same code unless it is required by the API.

example/opencv_demo.cc
```diff
@@ -133,10 +133,10 @@ int main(int argc, char *argv[])
...
 
         // Make an image_u8_t header for the Mat data
-        image_u8_t im = { .width = gray.cols,
-            .height = gray.rows,
-            .stride = gray.cols,
-            .buf = gray.data
+        image_u8_t im = { gray.cols,
+            gray.rows,
+            gray.cols,
+            gray.data
```
christian-rauch, line 139: Why is this change required?

CMakeLists.txt
```diff
@@ -37,17 +37,18 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
...
 add_library(${PROJECT_NAME} ${APRILTAG_SRCS} ${COMMON_SRC} ${TAG_FILES})
 
 if (MSVC)
+    target_compile_definitions(${PROJECT_NAME} PUBLIC "_CRT_SECURE_NO_WARNINGS")
     # FindThreads will not find pthread.h with MSVC
     # winmm is necessary for __imp_timeGetTime
-    find_library(PTHREAD_LIBRARIES NAMES pthreads)
-    target_link_libraries(${PROJECT_NAME} ${PTHREAD_LIBRARIES} winmm)
+    #find_library(PTHREAD_LIBRARIES NAMES pthreads)
+    #target_link_libraries(${PROJECT_NAME} ${PTHREAD_LIBRARIES} winmm)
```
christian-rauch, line 51: Remove commented code.

CMakeLists.txt
```diff
@@ -37,17 +37,18 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
 endif()
 
 aux_source_directory(common COMMON_SRC)
-set(APRILTAG_SRCS apriltag.c apriltag_pose.c apriltag_quad_thresh.c)
+set(APRILTAG_SRCS apriltag.c apriltag_pose.c apriltag_quad_thresh.c tagStandard41h12.c tag16h5.c tag25h9.c tag36h11.c tagCircle21h7.c tagCircle49h12.c tagCustom48h12.c tagStandard52h13.c)
```
christian-rauch, line 40: Why is this required?

christian-rauch (changes requested):
Thanks for your effort. I left some comments. On top of this, can you squash your commits so they appear more logically, e.g. one commit could add the new thread wrapper, another commit applies it to the code, another fixes other windows compile issues and the last one adds the CI integration.

I can look into the CI with ROS on Windows. This is probably not required. It's fine as long as we have one CI configuration for Windows to make sure that this does not break in future.

CMakeLists.txt
```diff
@@ -56,7 +57,7 @@ endif()
 set_target_properties(${PROJECT_NAME} PROPERTIES SOVERSION 3 VERSION ${PROJECT_VERSION})
 set_target_properties(${PROJECT_NAME} PROPERTIES DEBUG_POSTFIX "d")
 
-include(GNUInstallDirs)
+#include(GNUInstallDirs)
```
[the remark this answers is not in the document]
ryancwicks: it made no difference for me on Arch, but I'll leave it in.

common/pthreads_cross.cpp
```diff
@@ -0,0 +1,245 @@
+#include "common/pthreads_cross.h"
+#include <time.h>
```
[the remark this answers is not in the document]
ryancwicks: The original code is from here: https://nachtimwald.com/2019/04/05/cross-platform-thread-wrapper/

He has licensed any code on his site under an MIT license: https://nachtimwald.com/legal/

I've reproduced the license at the top of the header and cpp files, and have contacted the original author to double check with them.

common/pjpeg.c
ryancwicks: Fixed. You'll have to double check this for me. I don't use the C error handling routines very often.

example/apriltag_demo.c
ryancwicks: On older compilers, you couldn't set array sizes using const variable values, it needed to be a define. I've removed it since I don't think this has been an issue for many years.

example/opencv_demo.cc
ryancwicks: The direct initialization is a relatively new language feature that wasn't supported on older compilers. I've changed it, but it may not work on some older compilers.

example/apriltag_demo.c
ryancwicks: Nevermind, this needs to be a define to work with the Windows compiler. I've removed the const int declation, the #ifdefs and switched it to a #define.

christian-rauch (changes requested):
This looks good now. Thank you a lot!

My final request would just be to squash the commits into fewer commits, e.g. one for adding the new `pthread_cross` library, one for adapting the code base and one for the CI integration.

ryancwicks-voyis:
My attempt to squash the commits that far back in time seems to have broken things. I'll fix this later tonight by rolling back to the pre-rebase branch. It might be safer to leave the older commits as they are and I'll try to be cleaner with my commit messages in the future.

christian-rauch:
It looks like you merge the master branch into your PR. Did you maybe try to squash some of the commits that were previously merged into master, into your commits?

I think the safest thing to do is to rebase onto the current master, such that your PR only contains your own commits, and then squash your commits.

ryancwicks:
Sorry for the confused pull request. I've remerged my changes into another branch and created a new pull request with a cleaner commit history here https://github.com/AprilRobotics/apriltag/pull/234. The changes suggested in this pull request should reflected already in the new branch.

PR #103, opened 2020-08-17
Fix memory corruption on mem_pools in do_gradient_clusters() 

francis-maes:
Hello,

I found myself in a case where `nclustermap < mem_chunk_size` resulting in `sizeof(struct uint64_zarray_entry *)*2*nclustermap/mem_chunk_size == 6`, which is not a multiple of the size of pointer!

At least the code should be:
`malloc(sizeof(struct uint64_zarray_entry *)*(2*nclustermap/mem_chunk_size));`
in order to ensure we get an integer amount of pointer.

I made the following patch, since it appears that we at least need a size of 1:
`malloc(sizeof(struct uint64_zarray_entry *)*(1 + 2 * nclustermap / mem_chunk_size));`


mkrogius:
Thanks for the fix!

PR #106, opened 2020-09-20
Fix assumption that CMAKE_INSTALL_*DIR paths are relative

lopsided98:
The `CMAKE_INSTALL_*DIR` variables are documented as potentially being absolute paths. This PR allows apriltag to build with absolute installation paths. See https://github.com/jtojnar/cmake-snips#assuming-cmake_install_dir-is-relative-path for more information.

`target_include_directories()` already interprets the `$<INSTALL_INTERFACE:...>` expression relative to the installation prefix, so using `$<INSTALL_PREFIX>` is unnecessary, and breaks when `CMAKE_INSTALL_INCLUDEDIR` is an absolute path.

Setting `ARCHIVE_OUTPUT_DIRECTORY` and `LIBRARY_OUTPUT_DIRECTORY` breaks the build when `CMAKE_INSTALL_LIBDIR` is absolute. Removing that statement fixes the build and doesn't appear to change the installation behavior.

I have tested this PR with both relative and absolute `CMAKE_INSTALL_LIBDIR` and `CMAKE_INSTALL_INCLUDEDIR`.


lopsided98:
BTW, [ROS uses a patch](https://github.com/AprilRobotics/apriltag-release/blob/patches/release/noetic/apriltag/0001-Set-CMAKE_INSTALL_LIBDIR-to-lib.patch) that I think was needed as a workaround for this issue.

mkrogius:
Thanks for the PR!

PR #111, opened 2020-10-15
Improves build system for edge cases.

atuleu:
This pull request improves the build system on two points:
 * Uses PROJECT_SOURCE_DIR instead of CMAKE_SOURCE_DIR : When trying to build apriltag using cmake `ExternalProject_Add` or `FetchContent`, the python and tags macros fails, as `CMAKE_SOURCE_DIR` refers to the parent's project directory and not apriltag's project directory.
*  Makes the python wrapper build optional. Just add an option to hide any message warning in automated build system if one is not interested in python wrapper. The default behavior is conserved


mkrogius:
Thanks very much! I intended to do something like this but never got around to it

PR #112, opened 2020-10-22
Fix unnecessary escaping in CMake

jerry73204:
I found the `cmake && make` failed when CFLAGS contains whitespaces, such as `CFLAGS="-march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fstack-protector-strong"`. The cmake emits the command like this:

```
cc -march=x86-64\ -mtune=generic\ -O2\ -pipe\ -fstack-protector-strong\ -fno-plt ...
```

This patch fixes unnecessary whitespace escaping in CMake.

EDIT:
The command was wrong. It should be

```
export CPPFLAGS="-D_FORTIFY_SOURCE=2"
export CFLAGS="-march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fstack-protector-strong" 
cmake ../apriltag-3.1.3 -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr && make VERBOSE=1
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,7 +75,7 @@ if (NOT Python3_NOT_FOUND AND NOT Numpy_NOT_FOUND AND PYTHONLIBS_FOUND AND BUILD
 # TODO deal with both python2/3
 execute_process(COMMAND python3 ${PROJECT_SOURCE_DIR}/python_build_flags.py OUTPUT_VARIABLE PY_OUT)
 set(PY_VARS CFLAGS LDFLAGS LINKER EXT_SUFFIX)
-cmake_parse_arguments(PY "" "${PY_VARS}" "" ${PY_OUT})
+cmake_parse_arguments(PY "" ${PY_VARS} "" ${PY_OUT})
 separate_arguments(PY_CFLAGS)
 separate_arguments(PY_LDFLAGS)
 


mkrogius:
What OS are you using?

mkrogius:
This change actually breaks the build of the python library on Ubuntu.

jerry73204:
I'm using Arch Linux with cmake 3.16.3. I suspect the cmake version makes the difference.

mkrogius:
Are you certain that on your system, with your change, the apriltag python library is getting built? On my system with this change, there is no error, its just that the python library never gets produced. Check that after building, a file named something like apriltag.cpython-36m-x86_64-linux-gnu.so exists in your build folder.

jerry73204:
I did not notice that it omits the python build, and yes that patch would be invalid.
I realize the command in first post was not correct. I build the apriltag with makepkg, an automated process on Arch. The command was scraped from the config. Here is the command that actually fails on both Arch Linux and Ubuntu 20.04.

```sh
export CPPFLAGS="-D_FORTIFY_SOURCE=2"
export CFLAGS="-march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fstack-protector-strong" 
cmake ../apriltag-3.1.3 -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr && make VERBOSE=1
```

mkrogius:
Thanks for following up, sorry for the slow response.

I've tried it with these commands but for me it builds (on Ubuntu 18.04). This is what gets generated for the linker command for apriltag_demo:

/usr/bin/cc -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fstack-protector-strong -O3 -DNDEBUG -rdynamic CMakeFiles/apriltag_demo.dir/example/apriltag_demo.c.o -o apriltag_demo  -Wl,-rpath,/home/max/april/apriltag/build: libapriltag.so.3.1.0 -lpthread -lm

jerry73204:
Interesting. Ubuntu 18.04 does not work for me. Here is my screenshot of error and my commands on 3.1.3.
https://imgur.com/IB7MA2d

Can you check the cmake version? I guess yours is Ubuntu's default 3.10.2.

mkrogius:
I'm actually on cmake version 3.18.0. That could explain the differences although I'm not totally sure. The flags for this specific command are actually generated by the python_build_flags.py file. Can you run that file with python and see what flags it outputs? It might be that the fix is to change the python code in that file to process the newlines differently.

jerry73204:
This is what I get on Arch Linux with cmake 3.18.4.
https://imgur.com/0XL6p4f

mkrogius:
Hmm, so the flags output by the python script seem totally fine. The relevant line of CMakeLists is:

add_custom_command(OUTPUT apriltag_pywrap.o
    COMMAND ${CMAKE_C_COMPILER} ${PY_CFLAGS} -I${PROJECT_BINARY_DIR} -c -o apriltag_pywrap.o ${PROJECT_SOURCE_DIR}/apriltag_pywrap.c
    DEPENDS ${PROJECT_SOURCE_DIR}/apriltag_pywrap.c ${PROJECT_BINARY_DIR}/apriltag_detect.docstring.h ${PROJECT_BINARY_DIR}/apriltag_py_type.docstring.h)

The only thing that comes before ${PY_CFLAGS} is ${CMAKE_C_COMPILER}, so I think it is actually that variable that has gotten messed up for you. You could try explicitly setting this variable with:
    cmake -DCMAKE_C_COMPILER="/path/to/your/c/compiler/executable"

Alternately, deleting your build folder and recreating it should hopefully clear this error since its probably something that has just gotten stuck in your cache.

christian-rauch:
I cannot reproduce the issue with the commands you listed. I will close this for now.

PR #117, opened 2020-12-16
[cmake] install fixes and refactoring

### LINKED ISSUE #116: No version information in apriltagConfig.cmake makes find_package(apriltag <version> CONFIG) fail
Example cmake output when using find_package(apriltag 3.0 CONFIG):

```
Could not find a configuration file for package "apriltag" that is
  compatible with requested version "3.0".

  The following configuration files were considered but not accepted:

    /opt/AliceVision_install/share/apriltag/cmake/apriltagConfig.cmake, version: unknown
```
apriltag_FOUND will be false in this case.

simogasp:
Here are few fixes and refactoring of the CMakeLists file to make the export more standard (in terms of modern cmake).

* add version and languages to project() so the library version can be defined once and then used as ${PROJECT_VERSION}

* Added two options:
   * Standard BUILD_SHARED_LIBS (Default ON) to enable building shared or static libs
   * CMAKE_BUILD_TYPE will default to release if not specified

* removed `include_directory()`, its use is not recommended in modern cmake, better use per target includes.

* add a debug postfix when the library is built in debug (it adds the standard "d" at the end of the name)

* refactoring of the install/export of cmake config files, most notably
   * the config file is now parametric and generated automatically
   * a version file is generated to allow dealing with compatibility (fixes #116 )
   * all generated files are saved in the build directory under a "generated" folder
   * cmake config files are installed in lib/cmake rather than share (it is more "standard")

* [minor] fixed indentation for the python part

 PS
If i may, I'd suggest following a more standard organization of the repository, collecting all the code out of the root folder into a `src` subfolder or something like that. I can propose it in another PR in case.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,7 +32,7 @@ else()
     target_link_libraries(${PROJECT_NAME} PUBLIC Threads::Threads m)
 endif()
 
-set_target_properties(${PROJECT_NAME} PROPERTIES SOVERSION 3 VERSION 3.1.0)
+set_target_properties(${PROJECT_NAME} PROPERTIES SOVERSION 3 VERSION ${PROJECT_VERSION})
 
 include(GNUInstallDirs)
 target_include_directories(${PROJECT_NAME} PUBLIC
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,7 +38,6 @@ set_target_properties(${PROJECT_NAME} PROPERTIES DEBUG_POSTFIX "d")
 include(GNUInstallDirs)
 target_include_directories(${PROJECT_NAME} PUBLIC
     "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/>"
-    "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/common>"
     "$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>/apriltag")
 
 # install library
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,13 +3,21 @@ project(apriltag VERSION 3.1.0 LANGUAGES C CXX)
 
 option(BUILD_SHARED_LIBS "Build shared libraries" ON)
 
+# Set a default build type if none was specified
+set(default_build_type "Release")
+
+if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
+    message(STATUS "Setting build type to '${default_build_type}' as none was specified.")
+    set(CMAKE_BUILD_TYPE "${default_build_type}" CACHE  STRING "Choose the type of build." FORCE)
+    # Set the possible values of build type for cmake-gui
+    set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS  "Debug" "Release" "MinSizeRel" "RelWithDebInfo")
+endif()
+
 find_package(OpenCV QUIET)
 
 include_directories(.)
 aux_source_directory(common COMMON_SRC)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -45,58 +49,90 @@ foreach(HEADER ${HEADER_FILES})
...
+    add_custom_command(OUTPUT apriltag${PY_EXT_SUFFIX}
+        COMMAND ${PY_LINKER} ${PY_LDFLAGS} -Wl,-rpath,lib apriltag_pywrap.o $<TARGET_FILE:apriltag> -o apriltag${PY_EXT_SUFFIX}
+        DEPENDS ${PROJECT_NAME} apriltag_pywrap.o)
+    add_custom_target(apriltag_python ALL
+        DEPENDS apriltag${PY_EXT_SUFFIX})
+
+    execute_process(COMMAND python3 -m site --user-site OUTPUT_VARIABLE PY_DEST)
+    string(STRIP ${PY_DEST} PY_DEST)
+    install(CODE "execute_process(COMMAND cp ${PROJECT_BINARY_DIR}/apriltag${PY_EXT_SUFFIX} ${PY_DEST})")
 endif (NOT Python3_NOT_FOUND AND NOT Numpy_NOT_FOUND AND PYTHONLIBS_FOUND AND BUILD_PYTHON_WRAPPER)
```
fabiencastan, line 138: I would recommend to move all sample applications at the end, make it optional and move the find_package(opencv) in it:
```
# Examples

option(BUILD_DEMO "Build sample application" OFF)
option(BUILD_DEMO_OPENCV "Build sample application (using opencv)" OFF)

if(BUILD_DEMO)
# apriltag_demo
add_executable(apriltag_demo example/apriltag_demo.c)
target_link_libraries(apriltag_demo apriltag)
# install example programs
install(TARGETS apriltag_demo RUNTIME DESTINATION bin)
endif(BUILD_DEMO)

if(BUILD_DEMO_OPENCV)
find_package(OpenCV QUIET)
# opencv_demo
if(OpenCV_FOUND)
    add_executable(opencv_demo example/opencv_demo.cc)
    target_link_libraries(opencv_demo apriltag ${OpenCV_LIBRARIES})
    set_target_properties(opencv_demo PROPERTIES CXX_STANDARD 11)
    install(TARGETS opencv_demo RUNTIME DESTINATION bin)
endif(OpenCV_FOUND)
endif(BUILD_DEMO_OPENCV)
```

fabiencastan:
To integrate it to vcpkg (https://github.com/microsoft/vcpkg/issues/12171), we need to make the install more standard. Currently vcpkg generates the following errors:
```
The folder /include is empty or not present. This indicates the library was not correctly installed.
The software license must be available at ${CURRENT_PACKAGES_DIR}/share/apriltag/copyright
```

fabiencastan:
To recognize pthreads with vpkg, we also need to adjust:
```
if (MSVC)
    # FindThreads will not find pthread.h with MSVC
    # winmm is necessary for __imp_timeGetTime
    find_package(pthreads)
    if(pthreads_FOUND)
        # pthreads as structured in vcpkg
        target_link_libraries(${PROJECT_NAME} PRIVATE PThreads4W::PThreads4W winmm)
    else()
        find_library(PTHREAD_LIBRARIES NAMES pthreads)
        target_link_libraries(${PROJECT_NAME} ${PTHREAD_LIBRARIES} winmm)
    endif()
else()
    find_package(Threads REQUIRED)
    target_link_libraries(${PROJECT_NAME} PUBLIC Threads::Threads m)
endif()
```

mkrogius:
Thanks for the PR!

I made one small change of keeping the version info in the same location, just to avoid potentially breaking anyone's build.

Thanks for the offer to reorganize the repo, but I don't want reorganize at this time since I don't have that much time to spend on apriltag

PR #133, opened 2021-02-11
don't overwrite CMAKE_BUILD_TYPE

berndpfrommer:
Title says it all.
The overwriting of the cmake build type breaks my ROS2 build.
Note: this is fixed in PR #117 but that one has not seen progress in a while.
Please consider this PR, thanks!


berndpfrommer:
With the merging of PR #117 this PR is now obsolete.

PR #135, opened 2021-03-02
Revert change that removed an iteration limit in matd_svd_tall.

### LINKED ISSUE #134: Infinite loop
In my application of this library, I have come across an infinite loop (or at least very-long-running) when detecting tags. I have yet to be able to reproduce it in a context where I have the raw inputs logged, but I have encountered the issue roughly a dozen times.

CPU profile data pointed me to `matd_svd_tall`. That function spends a significant amount of time in the call stack `svd22 -> __cos_sse2 -> __dubsin` (which may help diagnose this as a platform-specific issue).

Reviewing the code, this comment caught my eye:
```
    // maxiters used to be smaller to prevent us from looping forever,
    // but this doesn't seem to happen any more with our more stable
    // svd22 implementation.
    int maxiters = 1UL << 30;
```

I have reverted the change to use the previous value (`int maxiters = 200 + 2 * A->nrows * A->ncols;`) and the issue has gone away in my application, with sufficient samples for me to be pretty confident.

I'm not sure what other side effects reverting this change could have, or if there's a more appropriate value to use, but I will send a PR in case others are encountering this issue from time to time.

kjkjava:
This fixes #134 in my application.

### DIFF
--- a/common/matd.c
+++ b/common/matd.c
@@ -1139,10 +1139,10 @@ static matd_svd_t matd_svd_tall(matd_t *A, int flags)
         }
     }
 
-    // maxiters used to be smaller to prevent us from looping forever,
-    // but this doesn't seem to happen any more with our more stable
-    // svd22 implementation.
-    int maxiters = 1UL << 30;
+    // empirically, we find a roughly linear worst-case number of iterations
+    // as a function of rows*cols. maxiters ~= 1.5*nrows*ncols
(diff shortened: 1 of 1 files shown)


mkrogius:
Thanks for the PR! I haven't actually ever had this issue crop up, so please send me an input image that reproduces the problem if you ever find one.

PR #140, opened 2021-03-22
[cmake] fix include path so that we can include <apriltag/apriltag.h>

simogasp:
The include directory when using apriltag as a 3rd party points to `install/path/include/apriltag/` which it is not what you really want if you want to include the headers using the prefix `apriltag` (to e.g. avoid potential name conflicts)
This fix also add `install/path/include/` to the search path so that we can `#include <apriltag/apriltag.hpp>`.
I kept the other to avoid breaking the compatibility with previous versions (i.e. so including `#include <apriltag.hpp>` still works but I'd not recommend itit).


fabiencastan:
It would be great to merge this PR.

This small issue is preventing us from integrating AprilTag into AliceVision:
https://github.com/alicevision/AliceVision/pull/950

Thanks for your time.

fabiencastan:
Any feedback?

mkrogius:
Apologies for the delay, I didn't really understand the change, but now I've read up more on this part of cmake.

fabiencastan:
Thanks!

PR #159, opened 2021-07-06
Resolve warnings; signed unsigned comparison

csaba-luxonis:
Resolved comparison between signed and unsigned.


christian-rauch:
Instead of converting the upper bounds of the loop to a signed "int", I would prefer if you convert the running variable to an unsigned int (`size_t` for memory sizes) to better indicate that this running variable is a positive index.

christian-rauch:
This is now superseded by PR https://github.com/AprilRobotics/apriltag/pull/309 where this and other warnings are fixed and reported as errors to make sure that they do not appear again.

PR #172, opened 2021-09-13
Optionally build static version

Robaire:
Respects the BUILD_SHARED_LIBS options to enable static builds if desired. Also disables the build of the python wrapper when building statically. This should remain compatible with any current build workflows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,9 @@
 cmake_minimum_required(VERSION 3.1)
 project(apriltag VERSION 3.1.0 LANGUAGES C CXX)
 
+if(POLICY CMP0077)
+    cmake_policy(SET CMP0077 NEW)
+endif()
 option(BUILD_SHARED_LIBS "Build shared libraries" ON)
 
 # Set a default build type if none was specified
@@ -18,7 +21,8 @@ set(APRILTAG_SRCS apriltag.c apriltag_pose.c apriltag_quad_thresh.c)
 
 # Library
 file(GLOB TAG_FILES ${PROJECT_SOURCE_DIR}/tag*.c)
-add_library(${PROJECT_NAME} SHARED ${APRILTAG_SRCS} ${COMMON_SRC} ${TAG_FILES})
+add_library(${PROJECT_NAME} ${APRILTAG_SRCS} ${COMMON_SRC} ${TAG_FILES})
+
 if (MSVC)
     # FindThreads will not find pthread.h with MSVC
     # winmm is necessary for __imp_timeGetTime
@@ -98,7 +102,8 @@ install(FILES "${PROJECT_BINARY_DIR}/apriltag.pc" DESTINATION "${CMAKE_INSTALL_L
 
 
 # Python wrapper
-option(BUILD_PYTHON_WRAPPER "Builds Python wrapper" On)
+include(CMakeDependentOption)
+cmake_dependent_option(BUILD_PYTHON_WRAPPER "Builds Python wrapper" ON BUILD_SHARED_LIBS OFF)
 
 if(BUILD_PYTHON_WRAPPER)
     SET(Python_ADDITIONAL_VERSIONS 3)


christian-rauch:
The change makes sense to me. This appears to be the same as https://github.com/AprilRobotics/apriltag/pull/137. Can we also just merge the old one or is there a reason for this PR?

Robaire:
They are identical and the old one should merge without issue. I resubmitted the PR since the other had been left stale for several months.

CMakeLists.txt
```diff
@@ -1,6 +1,9 @@
 cmake_minimum_required(VERSION 3.1)
 project(apriltag VERSION 3.1.0 LANGUAGES C CXX)
 
+if(POLICY CMP0077)
+    cmake_policy(SET CMP0077 NEW)
+endif()
```
christian-rauch, line 6: What is policy `CMP0077` needed for? Does this require a newer CMake version?

CMakeLists.txt
```diff
@@ -100,7 +104,8 @@ install(FILES "${PROJECT_BINARY_DIR}/apriltag.pc" DESTINATION "${CMAKE_INSTALL_L
 
 
 # Python wrapper
-option(BUILD_PYTHON_WRAPPER "Builds Python wrapper" On)
+include(CMakeDependentOption)
+cmake_dependent_option(BUILD_PYTHON_WRAPPER "Builds Python wrapper" ON BUILD_SHARED_LIBS OFF)
```
christian-rauch, line 108: Can the python module not be built with static linking?

christian-rauch (changes requested):

christian-rauch:
> I resubmitted the PR since the other had been left stale for several months.

I think in this case you can also just ping the maintainers or the people that recently merged stuff. This makes it easier to follow the PR discussions. In this particular case, it does not matter so much.

I left some comments. On top of this, can you:
1. squeeze some of your commits together so that you don't undo your own changes in subsequent commits? E.g. the second commit just undoes most of the first commit and the third commit just works together with the fourth commit,
2. rebase on master, so that your PR does not have merge commits?

CMakeLists.txt
Robaire: When I tried building the wrapper with a static library it failed. I didn't look into it too much because I don't know much about it and I assumed that someone building a static library intends to statically link it into a C executable. It wouldn't surprise me if the Python wrapper required a dynamic library to work, but I don't know for sure.

CMakeLists.txt
Robaire: It does not require a new version of CMake. It was to suppress a warning from CMake when using version 3.13 or greater, as the behavior of `option()` was changed slightly, as noted [here](https://cmake.org/cmake/help/latest/policy/CMP0077.html). Should the minimum CMake version ever be increased to 3.13 or higher this can be removed.

christian-rauch: But if `option` is removed in this PR, then we don't need this.

christian-rauch:
I think your rebase did not work. There are more commits in this PR than before (since you included some from master). Can you rebase your changes (and your changes only) on top of master (see https://docs.github.com/en/get-started/using-git/about-git-rebase)?

Robaire:
Fixed rebase issue.

christian-rauch:
@Robaire There is one comment left from my side (see the "change requested" hint): Do we still need to set the "CMP0077" since the "option" that it was added for, is removed with this PR?

christian-rauch (changes requested):
Clarify need for setting CMP0077 with remove "option".

CMakeLists.txt
Robaire: `option` is not removed in this PR. The behavior of `option` changed in CMake version 3.13. Prior to 3.13 `option` would disregard non-cached variables. 

For example, in 3.12 the following snippet would result in `BUILD_SHARED_LIBS` having a value of `ON`. In 3.13 and later it would have a value of `OFF`.

```cmake
set(BUILD_SHARED_LIBS OFF)
option(BUILD_SHARED_LIBS "Build shared libraries" ON)
```

Setting `CMP0077` indicates to CMake versions 3.13 and greater we explicitly intend to use the new behavior and not to issue a warning about it.

christian-rauch: I might be wrong here, but [`BUILD_SHARED_LIBS`](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html) is already a global CMake variable and specifying it again as `option` here is redundant. The condition for `BUILD_PYTHON_WRAPPER` should jus work with the global `BUILD_SHARED_LIBS` without redefining it.

Robaire: `BUILD_SHARED_LIBS` does not have a default value. If you omit `option` it will build statically by default. In order to maintain the default behavior of the library it has to be included. The [documentation](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html) states, "This variable is often added to projects as an option() so that each user of a project can decide if they want to build the project using shared or static libraries."

Robaire: You will only get a warning if `BUILD_SHARED_LIBS` is already defined as a non-cache variable. Add 'set(BUILD_SHARED_LIBS OFF)` directly above the option and you will see it.

Robaire: @simogasp Although ultimately I suppose the policy can also be set by the parent project that is linking against Apriltag. I'm not sure what the CMake standard approach is, to have each library set its policies or to rely on the parent setting policies?

@christian-rauch You are correct, it will build the static version if not defined (at least it did when I tested it, I don't know if the behavior is officially defined). Regardless of its behavior when not defined, it is typical to define it so there is always a specified default behavior.

christian-rauch (approved):

CMakeLists.txt
christian-rauch: Ok, in summary:
- we need the `option` to change the default behaviour to "shared"
- we need to set `CMP0077` to define the default `option` behaviour

I am happy with this.
@Robaire Can you maybe extend the commit message to document why setting `CMP0077` is required?

Robaire: @christian-rauch Done

christian-rauch: Thanks. And sorry for being pedantic about the `CMP0077` :-)

christian-rauch:
@mkrogius @wxmerkt I am happy with these changes. Can you let me know if I can go ahead and merge this?

christian-rauch:
@Robaire Can you rebase this once more? I am going to merge this now with a merge commit after your rebase.

Robaire:
@christian-rauch Can you explain what exactly you mean? I'm not very experienced with the more intricate git features. I used rebase to squash everything into one commit on top of the current upstream HEAD. Rebasing again right now doesn't seem to do anything, but rebase also has a lot of options that I may need to be using.

christian-rauch:
I meant doing a `git rebase origin/master` but I see now that you are already branching from the current `origin/master`. I also thought the last upstream commit was from two days ago, but now I see it's from the 27th of October.

So there is nothing to do for you. Sorry for the confusion.

Issue #174, opened 2021-09-24
Can not use the Tag Family in Python 3 after adding the .c and .h files and reinstalling


yusufguleray:
I would like to use the tag family `'tag36h10'` in my Python 3 program but even after adding the `'tag36h10.c'` and `'tag36h10.h'` and reinstalling with
```
$ cmake .
$ sudo make install
```
I get the
 
```
import apriltag
detector = apriltag.apriltag('tag36h10')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: Unrecognized tag family name: 'tag36h10'. Families I know about:
  tag36h11
  tag25h9
  tag16h5
  tagCircle21h7
  tagCircle49h12
  tagStandard41h12
  tagStandard52h13
  tagCustom48h12
```
How can fix this so that I can use 'tag36h10' in my Python 3 program?

Note: Here is the output of my installation
```
$ cmake .
-- The C compiler identification is GNU 7.5.0
-- The CXX compiler identification is GNU 7.5.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Configuring done
-- Generating done
-- Build files have been written to: /home/yus/MasterThesis/libraries/apriltag

$ sudo make install
Scanning dependencies of target apriltag
[  2%] Building C object CMakeFiles/apriltag.dir/apriltag.c.o
[  5%] Building C object CMakeFiles/apriltag.dir/apriltag_pose.c.o
[  7%] Building C object CMakeFiles/apriltag.dir/apriltag_quad_thresh.c.o
[ 10%] Building C object CMakeFiles/apriltag.dir/common/g2d.c.o
[ 13%] Building C object CMakeFiles/apriltag.dir/common/getopt.c.o
[ 15%] Building C object CMakeFiles/apriltag.dir/common/homography.c.o
[ 18%] Building C object CMakeFiles/apriltag.dir/common/image_u8.c.o
[...]

PR #175, opened 2021-09-27
Python wrapper change so that a new tag family can be used in Python

yusufguleray:
@christian-rauch Here is the PR for issue #174 

Building the repo with new tag families `.c` and `.h` files is not sufficient to use in Python. `apriltag_pywrap.c` also has to be updated accordingly.


christian-rauch:
I am fine with this change. @wxmerkt @mkrogius Any objections?

mkrogius (approved):
Looks good, thanks for the pr

Issue #182, opened 2021-11-28
Calls to exit() in apriltag.c


j-medland:
Hello

I have been developing a wrapper to allow for AprilTag to be integrated into other applications as a DLL/shared library.

During this development I have encountered a call to `exit()` which will terminate the whole application; an example being in the internal call `quick_decode_init` when adding a tag family to a detector.

I could overcome this in my code but would you be interested in changing this behaviour in your library?

I would be happy to contribute a PR but I don't know if you are interested or have an alternative approach in mind.

Cheers

John

mkrogius:
Hi,

I think that would be a good improvement to the library. Instead of `exit`ing, `quick_decode_init` and `apriltag_add_family_bits` could both return an error code to indicate success or failure. If you contribute a PR, I would review it.

Best,
Max

christian-rauch:
I also think this would be a nice contribution. A library should in general not control the exiting of the program.

j-medland:
Following discussion on #183 

I am happy to take a look at removing the `exit()` and `printf()` calls in the library.

With regards to `exit()`, I can find three calls:

1. `apriltag_detector_add_family*` calls can be modified from  a `void` to integer `return` which indicates success/failure when the internal `quick_decode_init` succeeds or fails.
2.  `workerpool_create` used by `apriltag_detector_detect` currently handles failure by exiting. The only way I could see to indicate that `apriltag_detector_detect` failed due to `workerpool_create` in a non-breaking way would be to return an empty detections `zarray` and set `errno` to something like `EAGAIN`
3.  `g2d_polygon_contains_point`  called by `apriltag_detector_detect` has the dead code branch which could just be removed
```c
    if (0 && v != g2d_polygon_contains_point_ref(poly, q)) {
        printf("FAILURE %d %d\n", v, quad_acc);
        exit(-1);
    }
```
With regards to `printf()` - most are easy to fix but;

1. Would you like to comment out the `printf` lines or should they be deleted?
2. `quick_decode_init` prints an error against hamming values over 3 but doesn't actually exit. I suggest we could return a failure from the function call and set `errno` to `EINVAL` or maybe `ERANGE`.

Let me know what you think

christian-rauch:
@j-medland Are you still interested in removing the `exit` calls in the library code? I am still interested in this and will be happy to review a PR that replaced those with error codes or similar.

christian-rauch:
Fixed by https://github.com/AprilRobotics/apriltag/pull/224.

PR #183, opened 2021-11-30
Remove exit from quick decode init

j-medland:
This PR aims to address issue #182.

I have defined the flag `APRILTAG_DETECTOR_ADD_FAMILY_OK` which is returned from  `apriltag_detector_add_family_bits` and `apriltag_detector_add_family` if the family provides preprocessor codes (`fam->impl`) or if `quick_decode_init` succeeds.

If the `add_family` functions are unsuccessful they return the number of bytes requested. This is handy for more informative error information but you might prefer to return just a fixed value.

I have modified the two examples - firstly to add the test for success and a call to `exit()` on failure and also to add the ability to specify the max hamming value for the `opencv_demo`.

I have also modified the python wrapper to check for success and return a (hopefully) meaningful error on failure.

I might consider these to be breaking changes to the API even though the original behaviour was non-ideal, but I shall leave that to you, the maintainers, to decide; I could extend the current API with a new call that adds the tag-family without calling exit and leave the external behaviour of the existing functions unchanged. Let me know if this would be preferred.

Quick tests of the functionality seem ok but I am happy to take another pass at anything and please feel free to make changes to the PR as you wish.

Cheers

John

### DIFF
--- a/apriltag.c
+++ b/apriltag.c
@@ -285,8 +285,8 @@ static size_t quick_decode_init(apriltag_family_t *family, int maxhamming)
         }
 
         printf("quick decode: longest run: %d, average run %.3f\n", longest_run, 1.0 * run_sum / run_count);
-        return APRILTAG_DETECTOR_ADD_FAMILY_OK;
     }
+    return APRILTAG_DETECTOR_ADD_FAMILY_OK;
 }
 
 // returns an entry with hamming set to 255 if no decode was found.
--- a/apriltag.c
+++ b/apriltag.c
@@ -196,7 +196,7 @@ static void quick_decode_uninit(apriltag_family_t *fam)
     fam->impl = NULL;
 }
 
-static void quick_decode_init(apriltag_family_t *family, int maxhamming)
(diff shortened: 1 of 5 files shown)


christian-rauch:
I personally prefer if the functions would just return static values, e.g. `0` on success and `-1` on failure. Further communication of the type of error could be done via `errno` (https://man7.org/linux/man-pages/man3/errno.3.html). Maybe we could reuse `ENOMEM` for the case when the allocation fails (I am not sure if this is already set by `calloc`). In any case, this behaviour should be documented.

Since you only change the return value from void to `size_t`, you are not really "breaking" API. Old code will still compile. I don't expect anyone relying on these functions to return void.

Since we are already cleaning up the library to make it more "compliant", we should also remove the calls to `printf` (or any other output functions). The communication between the library and the client/program should be solely by return values or status flags that can be read.

j-medland:
That all makes sense. I am going to close this and resume discussing it on #183 as there are some more decisions to discuss.

PR #186, opened 2021-12-14
switch to OpenCV4

christian-rauch:
This switches the Makefile to use OpenCV 4 and also sets the include path correctly via `pkg-config --cflags`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.1)
-project(apriltag VERSION 3.1.0 LANGUAGES C CXX)
+project(apriltag VERSION 3.2.0 LANGUAGES C CXX)
 
 if(POLICY CMP0077)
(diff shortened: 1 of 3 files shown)


wxmerkt (commented):
This would be correct on Ubuntu 20.04, and hence I am in favour of this change. However, I don't know when/in which distros OpenCV4 is the default OpenCV / when it will be installed in this directory (e.g. we recently ran into issues on Nvidia devkits that had non-standard OpenCV paths). That said, since most users will likely use CMake this should be fine.

mkrogius:
I also think this makes sense. We should bump the minor minor version number after this change though

mkrogius (approved):

christian-rauch:
I incremented the minor version 3.1 -> 3.2. Please tag this version after merge.

christian-rauch:
If this is ready to go through, can one of you (@mkrogius, @wxmerkt) merge this and tag a new release?

PR #187, opened 2021-12-14
update cmake documentation

christian-rauch:
- documents the use of `BUILD_SHARED_LIBS` for building either shared or static libraries
- optionally uses Ninja to build (it's much faster -- 2.2s (ninja) vs. 7.6s (make) on my system)
- uses `cmake --build build` to build build-system-agnostic
- build in separate `build` directory to not interfere with source files and the legacy `Makefile`

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,22 +25,38 @@ Usage
 Install
 =======
 
- Officially only linux operating systems are supported, although users have had success installing on windows too.
- 
- The default installation will place headers in /usr/local/include and
-shared library in /usr/local/lib. It also installs a pkg-config script
-into /usr/local/lib/pkgconfig and will install a python wrapper if python3 is installed.
+Officially only Linux operating systems are supported, although users have had success installing on Windows too.
 
-If you have CMake installed or it is not difficult to install, then do:
+The default installation will place headers in /usr/local/include and shared library in /usr/local/lib. It also installs a pkg-config script into /usr/local/lib/pkgconfig and will install a python wrapper if python3 is installed.
 
-    $ cmake .
-    $ sudo make install
-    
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,21 +25,29 @@ Usage
...
-into /usr/local/lib/pkgconfig and will install a python wrapper if python3 is installed.
-
-If you have CMake installed or it is not difficult to install, then do:
-
-    $ cmake .
-    $ sudo make install
-    
+The default installation will place headers in /usr/local/include and shared library in /usr/local/lib. It also installs a pkg-config script into /usr/local/lib/pkgconfig and will install a python wrapper if python3 is installed.
+
+## cmake
```
wxmerkt, line 32: Not everybody has ninja installed so it still makes sense for people to use `mkdir build ; cd build ; cmake .. -DCMAKE_BUILD_TYPE=Release ; make -j && make install`. We should recommend the use of ninja though, and then `cmake .. -GNinja -DCMAKE_BUILD_TYPE=Release ; ninja ; ninja install` should be enough (even though not pure CMake)

wxmerkt (approved):
Great, thank you for addressing this important question. I'd add a note that people can still use CMake + make rather than resorting to the Makefile directly

README.md
christian-rauch: Alright. I change the documentation back to using the default Makefile generator. The compilation via  `cmake --build build --target install` is still the same (it's the same for any generator).

wxmerkt (approved):

mkrogius:
Thank you for updating this

mkrogius (approved):

PR #188, opened 2021-12-15
Fix `refine_edges()` Corner Ordering

QuentinTorg:
The output quad corners of refine edges are rotated one index clockwise from the input quad corners. 

This is not a major issue, except when the intersection cannot be found, or the edge cannot be refined. In these error cases, the algorithm does not modify the original corner, but the original corner index is at the wrong index. 

When refining the edges, the edges are stored in an array. Edge 0 is the line from corners 0-1. Edge 1 is corners 1-2. 
This means that the intersection of edge 0 and edge 1 is corner 1. The current algorithm makes the assumption that the intersection of edge 0 and 1 will be corner 0. 

This PR fixes the ordering so that the output quad has the same corner indexes as the input quad, which also solves the ordering issues for the fallback cases

### DIFF
--- a/apriltag.c
+++ b/apriltag.c
@@ -868,8 +868,8 @@ static void refine_edges(apriltag_detector_t *td, image_u8_t *im_orig, struct qu
             double L0 = W00*B0 + W01*B1;
 
             // compute intersection
-            quad->p[i][0] = lines[i][0] + L0*A00;
-            quad->p[i][1] = lines[i][1] + L0*A10;
+            quad->p[(i+1)&3][0] = lines[i][0] + L0*A00;
+            quad->p[(i+1)&3][1] = lines[i][1] + L0*A10;
         } else {
             // this is a bad sign. We'll just keep the corner we had.
 //            printf("bad det: %15f %15f %15f %15f %15f\n", A00, A11, A10, A01, det);


christian-rauch:
> This is not a major issue, except when the intersection cannot be found, or the edge cannot be refined. In these error cases, the algorithm does not modify the original corner, but the original corner index is at the wrong index.

Do you have an example where this causes a wrong or inaccurate pose estimation?

mkrogius:
Seconding @christian-rauch, I don't see how this would end up causing a problem. The refine_edges step happens before quad_decode, which is where the final order of the corners is decided on

QuentinTorg:
> Do you have an example where this causes a wrong or inaccurate pose estimation?

I only have a theoretical failure case right now, but I can try to find or generate a real case eventually.

When everything works as planned, there is no issue. Rotating the points clockwise by one index will not cause any issues as long as all edges can be refined and all intersections are valid. When all works correctly, the input corners correspond to the output corners in the following way. 
```
quad_out->p[0]  derived from  quad_original->p[1] 
quad_out->p[1]  derived from  quad_original->p[2]
quad_out->p[2]  derived from  quad_original->p[3]
quad_out->p[3]  derived from  quad_original->p[0]
```
Quad decoding has no issue with the rotated corners as long as they stay ordered clockwise around the quad.


The problem happens if one of the corner intersection calculations fails. The failing corner index uses the original corner index, and not the newly rotated corner index. This leaves a duplicate corner in the quad. 

For the failure case, lets say the intersection of lines 0 and 1 cannot be computed. Line 0 is between original corners 0-1, and line 1 is between original corners 1-2. The intersection of lines 0-1 corresponds to the original corner 1. If the intersection cannot be computed though, then the intersection value is not inserted and the output quad will retain the value of original quad corner 0. The output quad will have the following mapping
```
quad_out->p[0]  is exactly    quad_original->p[0]
quad_out->p[1]  derived from  quad_original->p[2]
quad_out->p[2]  derived from  quad_original->p[3]
quad_out->p[3]  derived from  quad_original->p[0]
```


I think there is an argument that could be made that this failure should never happen. Maybe if the failure case happens, the quad decode will fail no matter what. It is tough to say without examples at this point. I see this as a parallel discussion to the correctness of this refinement algorithm though. This would instead open up a discussion if the refine_quad should instead return a failure and quad decode should be be skipped for a quad if the refinement fails

christian-rauch:
Without looking into the code, I would say that the best way to show that this issue exists is to share an example image (real or synthetic) that causes a wrong detection.

Is this an issue that you encountered on real or synthetic data or did you just find this while looking through the code?

QuentinTorg:
This is an issue that I found while testing segments of the code to better understand the edge refinement. I don't have any examples yet that will run all the way through the detection pipeline and fail quad decode because of this issue

QuentinTorg:
When refine_edges is true, detecting this image of a tagStandard52h13 tag will fail without my PR but work with my PR

detector settings: 
quad_sigma = 0   ***important
quad_decimate = 1
min_border_width = 5
max hamming = 0
decode_sharpening = 0.25

0_tagStandard52h13_failure

jrepp:
Without your change and with the recommended settings the apriltag_demo program produces a detection using your sample image. It's an interesting idea you're presenting - I'd like to try and reproduce your results. 

```
./apriltag_demo ./corner-ordering.jpg --decimate=1 --blur=0 --debug --iters=1 --threads=8 --family=tagStandard52h13
loading ./corner-ordering.jpg
image: ./corner-ordering.jpg 200x200
detection   0: id (52x13)-0   , hamming 0, margin  230.562
 0                             init        0.007000 ms        0.007000 ms
 1                       blur/sharp        0.002000 ms        0.009000 ms
 2                        threshold        0.944000 ms        0.953000 ms
 3                        unionfind        3.211000 ms        4.164000 ms
 4                    make clusters        3.797000 ms        7.961000 ms
 5            fit quads to clusters        6.420000 ms       14.381000 ms
 6                            quads        0.006000 ms       14.387000 ms
 7                decode+refinement        1.986000 ms       16.373000 ms
 8                        reconcile        0.004000 ms       16.377000 ms
 9                     debug output       13.777000 ms       30.154000 ms
10                          cleanup        0.007000 ms       30.161000 ms
hamm     1     0     0     0     0     0     0     0     0     0       30.154     7
Summary
hamm     1     0     0     0     0     0     0     0     0     0       30.154     7
```

christian-rauch:
@QuentinTorg I ran your example via `./apriltag_demo -f tagStandard52h13 --decimate=1 --blur=0 --debug --iters=1 ~/Downloads/2.jpg` (after converting from png to jpg via `convert`) and I am also getting a detection for `0: id (52x13)-0`.

Can you provide the actual file and `apriltag_demo` parameters that you used?

QuentinTorg:
I was working with a modified version of the opencv demo that allowed me to load a `.png` image directly. The jpeg compression is ruining the example by adding new noise. If I convert the .png image to .pgm instead, the example works. Github comments won't let me upload my exact .pgm example unfortunately, but I just made it using `convert corner-ordering.png corner-ordering.pgm`

Using the pgm, I get the following
```
$ ./apriltag_demo ./corner-ordering.pgm --decimate=1 --blur=0 --debug --iters=1 --threads=8 --family=tagStandard52h13 --refine-edges=false
loading ./corner-ordering.pgm
detection   0: id (52x13)-0   , hamming 0, margin  230.562
 0                             init        0.001000 ms        0.001000 ms
 1                       blur/sharp        0.001000 ms        0.002000 ms
 2                        threshold        0.245000 ms        0.247000 ms
 3                        unionfind        0.823000 ms        1.070000 ms
 4                    make clusters        0.603000 ms        1.673000 ms
 5            fit quads to clusters        3.621000 ms        5.294000 ms
 6                            quads        0.021000 ms        5.315000 ms
 7                decode+refinement        0.548000 ms        5.863000 ms
 8                        reconcile        0.001000 ms        5.864000 ms
 9                     debug output        6.420000 ms       12.284000 ms
10                          cleanup        0.001000 ms       12.285000 ms
hamm     1     0     0     0     0     0     0     0     0     0       12.284     7
Summary
hamm     1     0     0     0     0     0     0     0     0     0       12.284     7
```
```
$ ./apriltag_demo ./corner-ordering.pgm --decimate=1 --blur=0 --debug --iters=1 --threads=8 --family=tagStandard52h13 --refine-edges=true
loading ./corner-ordering.pgm
 0                             init        0.001000 ms        0.001000 ms
 1                       blur/sharp        0.001000 ms        0.002000 ms
 2                        threshold        0.245000 ms        0.247000 ms
 3                        unionfind        0.832000 ms        1.079000 ms
 4                    make clusters        0.612000 ms        1.691000 ms
 5            fit quads to clusters        3.596000 ms        5.287000 ms
 6                            quads        0.022000 ms        5.309000 ms
 7                decode+refinement        0.602000 ms        5.911000 ms
 8                        reconcile        0.001000 ms        5.912000 ms
 9                     debug output        6.631000 ms       12.543000 ms
10                          cleanup        0.000000 ms       12.543000 ms
hamm     0     0     0     0     0     0     0     0     0     0       12.542     7
Summary
hamm     0     0     0     0     0     0     0     0     0     0       12.542     7

```

christian-rauch:
> If I convert the .png image to .pgm instead, the example works.

Ok, I can reproduce this with the converted file. For reference, I am attaching the zip compressed pgm file:
[corner-ordering.zip](https://github.com/AprilRobotics/apriltag/files/7986927/corner-ordering.zip).

I still think this is a very synthetic example. It's virtually impossible to find this case in real data, but this might be an issue on synthetically rendered data.

If this doesn't affect the behaviour in other cases, I am fine with merging this. @mkrogius @wxmerkt What's your take on this? Can we merge this or do you see issues with this?

wxmerkt:
@christian-rauch I can't comment and haven't worked with this in detail. The example / case fix appears to work without a regression so likely okay to merge. Should we add a note into the code near the fix on why it's written the way it is (with reference or copy to the discussion in this PR). Otherwise `(i+1)%3` as an index leaves future readers to wonder "why"?.

mkrogius:
Apologies for the delay in responding. I went ahead and merged this fix, I think it makes a lot of sense theoretically. @QuentinTorg I'm curious how you managed to find this issue?

QuentinTorg:
@mkrogius I found this issue when trying to understand the performance/impact of the edge refinement algorithm on quad corner locations. I noticed that when comparing refined and unrefined quads their indexes would be different, and saw the root of the issue was this rotation of one index of the corners. 

I understood the failure theoretically and then created an image that would induce the failure case. This was not a failure that I discovered in the wild on real data. While possible, I think it would be quite difficult to find a real image that fails in this way

PR #197, opened 2022-01-25
Reuse p and R memory from params in apriltag_pose

Affie:
Hi, I want to update the Julia wrapper for apriltags and I have this change that fixes an issue. Memory not owned by the function is freed and new memory is allocated.
I don't know if this is the best way and I'm open to something better as long as the memory passed in is not freed in `matd_destroy`.

Thanks

### DIFF
--- a/apriltag_pose.c
+++ b/apriltag_pose.c
@@ -78,8 +78,10 @@ double orthogonal_iteration(matd_t** v, matd_t** p, matd_t** t, matd_t** R, int 
             matd_destroy(M2_update);
         }
         matd_scale_inplace(M2, 1.0/n_points);
-        matd_destroy(*t);
-        *t = matd_multiply(M1_inv, M2);
+        matd_t *tmp1 = matd_multiply(M1_inv, M2);
+        matd_set_data(*t, tmp1->data);
+        matd_destroy(tmp1);
+
         matd_destroy(M2);
 
         // Calculate rotation.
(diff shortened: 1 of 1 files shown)


christian-rauch:
What is the motivation behind this change? The memory for `t` is freed and allocated again. Is this an issue? Alternatively, you are allocating and freeing `tmp1`, which probably has the same cost.

Affie:
In the julia wrapper the memory is managed by julia and freeing it in the function causes a segfault. 
So it's not about the cost.

Affie:
Note, I call `orthogonal_iteration` directly to save on converting between julia and c types.

Affie:
Also note: we build binaries for easy installation and use with Julia https://github.com/JuliaBinaryWrappers/AprilTags_jll.jl/releases/tag/AprilTags-v0.1.0%2B0, it is built for all supported platforms, but I haven't tested it on other platforms. I would like to make maintenance easier and directly build from this repository, it's currently built from a fork.

christian-rauch:
@mkrogius @wxmerkt If you don't know of a better way to set the matrix memory directly without creating another matrix (`tmp1` and `tmp2`), I would go ahead with this proposal.

wxmerkt:
@christian-rauch I am afraid I can't comment on this. @mkrogius is the best person to answer

mkrogius:
I'm leaning against accepting this change because somebody reading this code will not understand why this code is written this way. Surely there must be some way in the Julia wrapper to create and pass in a "t" that is a native c type and so would not segfault?

PR #206, opened 2022-02-22
Removing execute bits from tag36h10.*

jrepp:


christian-rauch:
This makes sense. Not sure why these files got the executable bit set in the first place.

PR #208, opened 2022-02-22
Remove unused includes

jrepp:


christian-rauch:
Are those header files "unused" because they are already included by other header files or are their functions not used at all? Do you have a tool to provide this information or did you find this by manually removing the includes?

jrepp:
I have support in my IDE for [include-what-you-use](https://include-what-you-use.org/) - I could investigate integrating this into the build later but it seems fairly low priority compared to other items.

I try to abide by these guidelines:

All headers should include all of their dependencies
- this aids in portability and prevents playing whack-a-mole chasing down dependencies 

Source files should include their local dependencies before system dependencies
- This provides a better early signal on the above and allows tools like include-what-you-use to work better and give you minimal dependency coverage.

Headers should not expose unnecessary details
- Kinda obvious I think apriltag does a decent job here but we can actually remove visibility of the detector internals and zarray internals from library users.

christian-rauch:
Ok, this seems fine to me. It only touches .c files so there will be no impact on the public API from includes.

PR #209, opened 2022-02-22
Removing string copy from timeprofile calls

jrepp:


common/timeprofile.h
```diff
@@ -73,12 +73,12 @@ static inline void timeprofile_clear(timeprofile_t *tp)
...
 }
 
+// Caller owns name memory, use static memory: timeprofile_stamp(tp, "profile-section"); 
 static inline void timeprofile_stamp(timeprofile_t *tp, const char *name)
 {
     struct timeprofile_entry tpe;
 
-    strncpy(tpe.name, name, sizeof(tpe.name));
-    tpe.name[sizeof(tpe.name)-1] = 0;
+    tpe.name = name;
```
christian-rauch, line 81: If the caller owns the memory, how do we make sure it is not free'd before `tp->stamps` is used?

jrepp: It is always static memory passed to an internal API e.g. the memory lives inside the .text segment of the library and has the same lifetime scope as the functions accessing it:

image

mkrogius: I agree with @christian-rauch. Even if currently there are no uses of timeprofile_stamp which would cause a problem, I'd rather not create a footgun.

Was the goal of this change to speed up detection? If so, I would be willing to accept this change if it gives a substantial speedup

jrepp:
Desire is to remove measurement time from measurement API but given its infrequent calls in relation to the measured code this seems unnecessary. Thank you for reviewing!

Issue #210, opened 2022-02-23
Can we deprecate the Makefile?


jrepp:
There is further drift been make and cmake versions of the build. CMake is the default working version. If this is agreeable I can open a PR against this issue to fold the Makefile options into the CMake ensuring that we get apples-to-apples build comparison between make and a default cmake build.

jrepp:
I'm seeing this as a long-term tax for little benefit. The Makefile creates some non-supported surface area on the project. I'm hoping to add more testing and coverage support through CMake but don't plan to duplicate that work.

I'm ok with closing this issue - I just wanted to raise it as an issue for discussion.

mkrogius:
I think that makes a lot of sense and that there is no need to duplicate that work for the Makefile

acshi:
As far as this project is currently owned by the April Robotics Lab, I think we should acknowledge that the lab and alums from the lab have very little time to maintain this. So I think the default here should be to stay in a low-cost maintenance mode.

Regardless of that, I think that any changes should be heavily justified, either because of explicit examples where undefined behavior is invoked, or significant performance differences with the benchmarking numbers to prove them.

Glancing at recent issues, for example the diagnostics PR that seems to motivate this, I am not sure the use case for the changes is strongly made. Is there any performance difference? Is some specific use case not possible? The example given is specifying an invalid tag family and getting an assertion abort, which even if not the prettiest behavior, doesn't seem to block any specific use of AprilTag. So I would tentatively say that large-scale refactoring like that is perhaps better suited to a fork of the project... I haven't been very active looking at issues here, but I agree that I think performance and correctness are the main areas of concern, and so I think changes should be justified in those terms.

jrepp:
I am actively working on an AprilTag based solution for an industrial application for the remainder of this year and probably beyond. I have approval to upstream changes which is my preference so that the broader community can benefit from this work and we don't curate a dead end of an important piece of software. 

I'd like to keep contributing and am willing to offer my time as a maintainer if the team is amiable to it. I have experience in maintaining long-lived stable APIs (I was an Xbox platform developer) and doing low-level engine code for video games at companies like Blizzard, MS, and EA. Not to argue from authority just to say I have a lot of experience in writing, maintaining, and optimizing tricky performance-critical code.

Re: Makefile/CMake conversation:
[...]

PR #211, opened 2022-02-23
Support for disabling asserts at compile time

jrepp:
First support patch for the diagnostics changes in previous PR.

Some notes for implementors:

* apriltag_config.h is only needed for internal compilation units
* apriltag_config.h is included in apriltag_demo.c because for internal testing (similar to other includes in _demo.c)
* asserts currently print to stderr but I have another commit which makes this configurable at compile time
* I also have some other downstream changes for this that address a few issues such as changing exit/panic behavior etc at compile time.

This change only includes the change from assert to AT_ASSERT for compile-time configuration of asserts (including disabling them).

Testing

```
$ mkdir build && cd build
$ cmake -G Ninja ..
$ ninja
$ ./apriltag_demo -f foo
Unrecognized tag family name. Use e.g. "tag36h11".
apriltag_demo: /home/jrepp/dev/apriltag/example/apriltag_demo.c:105: main: Assertion `0' failed.
Aborted (core dumped)

$ ccmake ..
(toggle assert off)
$ ninja
$ ./apriltag_demo -f foo
$ ./apriltag_demo -f foo
Unrecognized tag family name. Use e.g. "tag36h11".
$ echo $?
255
```


christian-rauch:
These two commits should be squashed because the second commit fixes something from the first commit.

This PR also touches a lot of unrelated parts. There are a lot of changes in the includes and reordering in the `CMakeLists.txt`. It appears to me that these changes are also unrelated to the main purpose of this PR, which is changing the assertion behaviour. Those assertion-unrelated parts should be reviewed independently in a dedicated PR.

The assertions should be disabled in Release mode by default. So, if this is only about disabling assertions, then building in `Release` mode would be the right way. Or am I missing something here?

How is `AT_ASSERT` implemented in this PR? It is not obvious to me.

I think this will also break the make script (`Makefile`). We either have to deprecate and remove the Makefile (https://github.com/AprilRobotics/apriltag/issues/210), or we have to support it.

jrepp:
> These two commits should be squashed because the second commit fixes something from the first commit.

I get what you're saying about always having a single commit - but from the point of the PR or people reviewing the PR that's not what you usually want. 

Are you concerned with the commit history of the main branch? If that's the case you can't you just use the squash merge option when accepting the PR?

For example, if you request a change to the PR and I make said change, I shouldn't rewrite that commit, or the review history is borked and anyone trying to follow along or reproduce my results can no longer rebase.

That said I'm happy to conform to whatever makes you most comfortable. 

> The assertions should be disabled in Release mode by default. So, if this is only about disabling assertions, then building in Release mode would be the right way. Or am I missing something here?

I need a bit more flexibility than that as I'm compiling to different embedded platforms. On embedded platforms, we need the ability to create different flavor builds with and without asserts - release builds with asserts are really useful for testing different code generation options (optimization levels etc). 

Additionally, we need to control both assert and panic behavior and I can't defer that to the standard library. For example, in a follow-up commit, I will be able to close this issue: https://github.com/AprilRobotics/apriltag/issues/182 as we will be able to replace exit() and print with some compile-time configuration macros in diagnostic.h (controlled through apriltag_config.h)

common/diagnostic.h
```diff
@@ -0,0 +1,18 @@
...
+//
+
+
+//
+// Enable asserts which can terminate the program when exceptional conditions are encountered
+//
+#if defined(AT_DIAG_ENABLE_ASSERT)
+    #include <assert.h>
+    #define AT_ASSERT(expr) assert(expr)
+    #define AT_ASSERT_MSG(expr, format, ... ) if (!(expr)) { fprintf(stderr, format , ##__VA_ARGS__); assert(0); }
```
jrepp, line 14: @christian-rauch here's the implementation of AT_ASSERT - it is just defined to assert(expr) in the case where it's enabled or empty otherwise.

Makefile
```diff
@@ -23,6 +23,9 @@ install: libapriltag.so
 	@./install.sh $(PREFIX)/include/apriltag $(APRILTAG_HEADERS)
 	@ldconfig
 
+apriltag_config.h: apriltag_config.h.in
```
jrepp, line 26: @christian-rauch this generates a default apriltag_config.h (all #cmakefine lines will be define 1 lines) the default behavior for release builds of assert() will be unchanged. 

I don't think supporting a Makefile long term is sustainable. We should :hocho: it soon :)

common/diagnostic.h
christian-rauch: So `AT_ASSERT` is just implemented by `assert`. But is there a benefit of enabeling/disabeling this manually? Assertions will be disabled in Release mode anyway and having them enabled in Debug mode makes a lot of sense.

christian-rauch:
@jrepp Are you still interested in this? Looking at the proposed changes and the discussion, you should clearly split this PR into the assertion-related parts and those that just change the code style, fix typos, etc.

Regarding the assertions, I am still not convinced that what you propose is necessary. If you really need assertions in Release mode, why don't you define the corresponding defines? I think `NDEBUG` is responsible for enabling/disabling the `assert` macro. Otherwise, you could just redefine the `assert` macro for release builds when a certain CMake option is provided.

jrepp:
I'll 

> @jrepp Are you still interested in this? Looking at the proposed changes and the discussion, you should clearly split this PR into the assertion-related parts and those that just change the code style, fix typos, etc.

agreed

> Regarding the assertions, I am still not convinced that what you propose is necessary. If you really need assertions in Release mode, why don't you define the corresponding defines? I think `NDEBUG` is responsible for enabling/disabling the `assert` macro. Otherwise, you could just redefine the `assert` macro for release builds when a certain CMake option is provided.

That's definitely a simple and existing solution. I don't see anything wrong with that.

The reason I often break out defines under flags like this is to allow flexibility. Globally then you can either hide this flag behind NDEBUG or you can select it in a target. I like to have a release debug build as well as a debug profiled build for example.

I'm going to close this PR :)

PR #215, opened 2022-02-26
[cmake] sanitazier flags in C_FLAGS instead of CXX

simogasp:
Building with sanitizers is not currently working because the code is C and the flags are set for CXX instead.
Just need to use `CMAKE_C_FLAGS_DEBUG`


christian-rauch:
Whoops. That's a copy&paste oversight from my side. No wonder it never reported memory issues. Thanks for noticing this.

PR #216, opened 2022-02-28
add documentation on how to enable and use the AddressSanitizer

christian-rauch:
Thanks to https://github.com/AprilRobotics/apriltag/pull/215, the `AddressSanitizer` feature now works properly. I added documentation on how to enable and use it. Using these instructions, the `opencv_demo` example already reports numerous `runtime error: division by zero` errors.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,19 @@ To install to a different directory than /usr/local:
 
     $ PREFIX=/some/path sudo make install
 
+Debugging
+=========
+
+You can enable [AddressSanitizer](https://clang.llvm.org/docs/AddressSanitizer.html) to debug memory issues for Debug builds by setting the `ASAN` option:
+```
+cmake -B build -GNinja -DCMAKE_BUILD_TYPE=Debug -DASAN=ON
+cmake --build build
+```
+
+Mostly you can then run your executables as usual and inspect the sanitiser output. If you get a message like `ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.` you have to preload the corresponding `libasan.so.5` like this:
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,6 +61,19 @@ To install to a different directory than /usr/local:
 
     $ PREFIX=/some/path sudo make install
 
+Debugging
+=========
+
+You can enable [AddressSanitizer](https://clang.llvm.org/docs/AddressSanitizer.html) to debug memory issues for Debug builds by setting the `ASAN` option:
```
jrepp, line 67: Probably good for another commit but we should get in the practice of namespacing CMake options. The reason is that it's common practice to include a library CMakeLists.txt into larger projects but options live a global scope. I have been using the AT_ prefix but am happy to use something else.

README.md
```diff
@@ -61,6 +61,19 @@ To install to a different directory than /usr/local:
...
+```
+cmake -B build -GNinja -DCMAKE_BUILD_TYPE=Debug -DASAN=ON
+cmake --build build
+```
+
+Mostly you can then run your executables as usual and inspect the sanitiser output. If you get a message like `ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.` you have to preload the corresponding `libasan.so.5` like this:
```
jrepp, line 73: nit: sanitizer

christian-rauch: Well, depends on which side of the Atlantic you are living :-)

README.md
christian-rauch: Since this is only used in Debug builds and not in production, I would not worry too much about it. Do you think this can conflict with other CMake projects and their options?

christian-rauch:
I am going to merge this. @jrepp tested this to find issues with the string hash function.

PR #219, opened 2022-03-01
Fixing zhash/ASAN errors, improve hash

jrepp:
```../common/zhash.c:545:22: runtime error: left shift of 211382645 by 7 places cannot be represented in type 'int'```

Fixed the ASAN error
Noticed the hash was lacking so improved the string hash function to use an optimized [FNV](https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function) - it's a few more hashes and a multiply.

Currently, this str hash is only used to index command-line arguments.

### DIFF
--- a/common/zhash.c
+++ b/common/zhash.c
@@ -540,14 +540,15 @@ uint32_t zhash_str_hash(const void *_a)
 
     char *a = * (char**)_a;
 
-    int32_t hash = 0;
+    uint32_t hash = 0;
     while (*a != 0) {
-        hash = (hash << 7) + (hash >> 23);
-        hash += *a;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: common/zhash.c [hunk 1]
```diff
@@ -540,10 +540,11 @@ uint32_t zhash_str_hash(const void *_a)
 
     char *a = * (char**)_a;
 
-    int32_t hash = 0;
+    uint32_t hash = 0;
```
christian-rauch, line 543: In this case, we can skip the `(uint32_t) hash` cast later.

jrepp: oops, ty

PR #223, opened 2022-03-21
Update Detector Parameter Descriptions

QuentinTorg:
This PR updates the comments about parameters in the apriltag detector to give the user a better understanding of valid ranges and default values

### DIFF
--- a/apriltag.h
+++ b/apriltag.h
@@ -147,7 +147,8 @@ struct apriltag_detector
     // What Gaussian blur should be applied to the segmented image
     // (used for quad detection?)  Parameter is the standard deviation
     // in pixels.  Very noisy images benefit from non-zero values (e.g. 0.8).
-    // Gaussian blur kernel size is (int)(4*sigma). Add 1 if even
+    // Gaussian blur kernel size is (int)(4*fabssigma)). Add 1 if even.
+    // Negaive values will sharpen the image instead of blur
     // Sample kernel sizes are below:
     //
     // max sigma          kernel size
--- a/apriltag.h
+++ b/apriltag.h
@@ -130,39 +130,56 @@ struct apriltag_detector
     // User-configurable parameters.
 
     // How many threads should be used?
+    //
+    // Default: 1
     int nthreads;
 
     // detection of quads can be done on a lower-resolution image,
     // improving speed at a cost of pose accuracy and a slight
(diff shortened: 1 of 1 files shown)


apriltag.h
```diff
@@ -130,39 +130,56 @@ struct apriltag_detector
...
     float quad_sigma;
 
     // When non-zero, the edges of the each quad are adjusted to "snap
     // to" strong gradients nearby. This is useful when decimation is
     // employed, as it can increase the quality of the initial quad
     // estimate substantially. Generally recommended to be on (1).
+    // Very computationally inexpensive
     //
-    // Very computationally inexpensive. Option is ignored if
-    // quad_decimate = 1.
```
QuentinTorg, line 153: Option is no longer ignored if quad decimate is 1: https://github.com/AprilRobotics/apriltag/blob/master/apriltag.c#L895

apriltag.h
```diff
@@ -130,39 +130,57 @@ struct apriltag_detector
...
+    // in pixels.  Very noisy images benefit from non-zero values (e.g. 0.8).
+    // Gaussian blur kernel size is (int)(4*fabssigma)). Add 1 if even.
+    // Negaive values will sharpen the image instead of blur
+    // Sample kernel sizes are below:
+    //
+    // max sigma          kernel size
+    // 0.0 -> 0.499       1  (disabled)
+    // 0.5 -> 0.999       3
+    // 1.0 -> 1.499       5
+    // 1.5 -> 1.999       7
```
christian-rauch, line 158: This table is a bit confusing. You are referring to the look-up table for equation "kernel_size = 4 * abs(quad_sigma)":
https://github.com/AprilRobotics/apriltag/blob/bee5efb1cc263ec1459bca46193e3428220d52ba/apriltag.c#L1020-L1027

But why this mapping from 0 to 0.499, 0.5 to 0.999, etc.? That looks like 0.0 gives you 0.499 and 0.5 gives you 0.999. Do you mean that all values in this range map to the same kernel size?

The notation is a bit confusing and I would personally prefer if this is just properly formalised. With an ASCII or LaTeX function.

christian-rauch (changes requested):
Please merge your fixup commits. I.e. you have two commits that change the same part twice. Merge them to make the history more readable.

christian-rauch:
@QuentinTorg Are you interested in continuing with this?

PR #224, opened 2022-04-19
Replaced calls to exit()

j-medland:
Here is a belated PR for issue #182.

I have removed all the calls to `exit()` that might be encountered and set the `errno` where it isn't already set by the operation that failed.

As the user might not be checking the `errno` values I thought it best to avoid their application crashing by returning an empty detections array, where relevant, to avoid them encountering a null-pointer.

Any feedback is welcome and I am happy to tweak as required.

### DIFF
--- a/apriltag.c
+++ b/apriltag.c
@@ -37,6 +37,7 @@ either expressed or implied, of the Regents of The University of Michigan.
 #include <stdint.h>
 #include <string.h>
 #include <stdio.h>
+#include <errno.h>
 
 #include "common/image_u8.h"
 #include "common/image_u8x3.h"
@@ -46,6 +47,7 @@ either expressed or implied, of the Regents of The University of Michigan.
 #include "common/timeprofile.h"
 #include "common/math_util.h"
 #include "common/g2d.h"
+#include "common/debug_print.h"
 
 #include "apriltag_math.h"
 
@@ -214,13 +216,14 @@ static void quick_decode_init(apriltag_family_t *family, int maxhamming)
 
     qd->nentries = capacity * 3;
 
-//    printf("capacity %d, size: %.0f kB\n",
+//    debug_print("capacity %d, size: %.0f kB\n",
 //           capacity, qd->nentries * sizeof(struct quick_decode_entry) / 1024.0);
 
     qd->entries = calloc(qd->nentries, sizeof(struct quick_decode_entry));
     if (qd->entries == NULL) {
-        printf("apriltag.c: failed to allocate hamming decode table. Reduce max hamming size.\n");
-        exit(-1);
+        debug_print("Failed to allocate hamming decode table\n");
+        // errno already set to ENOMEM (Error No MEMory) by calloc() failure
+        return;
     }
 
     for (int i = 0; i < qd->nentries; i++)
@@ -254,13 +257,16 @@ static void quick_decode_init(apriltag_family_t *family, int maxhamming)
         }
 
         if (maxhamming > 3) {
-            printf("apriltag.c: maxhamming beyond 3 not supported\n");
+            debug_print("\"maxhamming\" beyond 3 not supported\n");
+            // set errno to Error INvalid VALue
+            errno = EINVAL;
+            return;
         }
     }
 
     family->impl = qd;
 
-    if (0) {
+    #if 0
         int longest_run = 0;
         int run = 0;
         int run_sum = 0;
@@ -282,7 +288,7 @@ static void quick_decode_init(apriltag_family_t *family, int maxhamming)
         }
 
         printf("quick decode: longest run: %d, average run %.3f\n", longest_run, 1.0 * run_sum / run_count);
-    }
+    #endif
 }
 
 // returns an entry with hamming set to 255 if no decode was found.
@@ -291,7 +297,8 @@ static void quick_decode_codeword(apriltag_family_t *tf, uint64_t rcode,
 {
     struct quick_decode *qd = (struct quick_decode*) tf->impl;
 
-    for (int ridx = 0; ridx < 4; ridx++) {
+    // qd might be null if detector_add_family_bits() failed
+    for (int ridx = 0; qd!=NULL && ridx < 4; ridx++) {
 
         for (int bucket = rcode % qd->nentries;
              qd->entries[bucket].rcode != UINT64_MAX;
@@ -440,7 +447,7 @@ static matd_t* homography_compute2(double c[4][4]) {
         }
 
         if (max_val < epsilon) {
-            fprintf(stderr, "WRN: Matrix is singular.\n");
+            debug_print("WRN: Matrix is singular.\n");
             return NULL;
         }
 
@@ -874,7 +881,7 @@ static void refine_edges(apriltag_detector_t *td, image_u8_t *im_orig, struct qu
             quad->p[(i+1)&3][1] = lines[i][1] + L0*A10;
         } else {
             // this is a bad sign. We'll just keep the corner we had.
-//            printf("bad det: %15f %15f %15f %15f %15f\n", A00, A11, A10, A01, det);
+//            debug_print("bad det: %15f %15f %15f %15f %15f\n", A00, A11, A10, A01, det);
         }
     }
 }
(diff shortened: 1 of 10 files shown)


apriltag.c
```diff
@@ -291,11 +296,12 @@ static void quick_decode_codeword(apriltag_family_t *tf, uint64_t rcode,
...
 
-    for (int ridx = 0; ridx < 4; ridx++) {
+    // quick_decode might be NULL if detector_add_family_bits() failed
+    for (int ridx = 0; qd!=NULL && ridx < 4; ridx++) {
 
         for (int bucket = rcode % qd->nentries;
-             qd->entries[bucket].rcode != UINT64_MAX;
-             bucket = (bucket + 1) % qd->nentries) {
+            qd->entries[bucket].rcode != UINT64_MAX;
+            bucket = (bucket + 1) % qd->nentries) {
```
christian-rauch, line 304: Can you move these "style changes" into a separate commit?

apriltag.c
```diff
@@ -306,7 +312,6 @@ static void quick_decode_codeword(apriltag_family_t *tf, uint64_t rcode,
 
         rcode = rotate90(rcode, tf->nbits);
     }
-
```
christian-rauch, line 309: Same here.. move into a dedicated commit for things that are not related to the removal of `exit()`.

apriltag.c
```diff
@@ -1001,6 +1006,12 @@ zarray_t *apriltag_detector_detect(apriltag_detector_t *td, image_u8_t *im_orig)
     if (td->wp == NULL || td->nthreads != workerpool_get_nthreads(td->wp)) {
         workerpool_destroy(td->wp);
         td->wp = workerpool_create(td->nthreads);
+        if(td->wp == NULL){
+            // creating workerpool failed
+            printf("apriltag.c: Creating workerpool failed.");
```
christian-rauch, line 1011: Is it possible to not print text on stdout in the core library? This should be controlled by a logger that can be turned on/off. Maybe we can also use `perror` as in the case below.

What are the reasons for failing to create a worker pool? Can we forward this error to the user via the `errno`?

common/g2d.c
```diff
@@ -404,6 +404,7 @@ int g2d_polygon_contains_point(const zarray_t *poly, double q[2])
     int v = (quad_acc >= 2) || (quad_acc <= -2);
 
     if (0 && v != g2d_polygon_contains_point_ref(poly, q)) {
+        // disabled - used for debug only
         printf("FAILURE %d %d\n", v, quad_acc);
         exit(-1);
     }
```
christian-rauch, line 410: If this is only for debugging and cannot be controlled at runtime, then this should be guarded by an #ifdef so that this "dead code" does not get compiled and we can skip the if condition.

Better, this should probably be an `assert` so that it gets automatically tested in debug mode but not compiled into release builds.

I know, it's not related to the problem you try to solve but if you are in the mood, you could still add a commit for this :-)

common/workerpool.c
```diff
@@ -115,7 +116,8 @@ workerpool_t *workerpool_create(int nthreads)
             int res = pthread_create(&wp->threads[i], NULL, worker_thread, wp);
             if (res != 0) {
                 perror("pthread_create");
-                exit(-1);
+                // errno already set to EAGAIN by pthread_create() failure
+                return NULL;
```
christian-rauch, line 120: This is still printing on the terminal via [`perror`](https://en.cppreference.com/w/c/io/perror). Do we want the library to print to stderr?

christian-rauch (changes requested):
Thanks for this. I left some comments about printing on the terminal. Can you also split the "code style" changes that are unrelated to the main problem into a separate commit?

j-medland:
I am happy to remove the style changes. I thought I had caught them all prior to submitting the PR.

With regards to printing messages to the console - An alternative to using `printf()` would be good.

What error strategy would you prefer?

1. Just use `errno`
2. Write all information to `stderr` using `perror` or some debug-print macro that is disabled when building the library for release
3. Same as 2 but write to a text-file.
4. Have some global error store that stores the previous 1/5/10/n errors as strings with some API calls to query it.
5. Something else

Cheers

apriltag.c
j-medland: The call to `pthread_create` in `workerpool_create` can fail if the system cannot provide all the threads requested (50,000 for instance) - in doing so it sets the `errno` to `EAGAIN`

So checking for `EAGAIN` after `apriltag_detector_detect()` should show that something has gone wrong.

christian-rauch: If `errno` is set already, can we skip the `printf`? I see that there are many `printf` in other places, but I think those should also be replaced in the long-term with something the user can control.

I suggest that we use `errno` to indicate errors and `perror` if required to print on `stderr`. I would avoid printing to `stdout`.

j-medland:
I have made some updates which try to unify the printing of debug messages with the addition of a `debug_print` macro which will be removed in release builds and added in error checking to the examples in the relevant locations.

```bash
$ apriltag_demo -a 3 -f tagStandard41h12 tag.png
apriltag.c:224:quick_decode_init(): Failed to allocate hamming decode table
Unable to add family to detector due to insufficient memory to allocate the tag-family decoder. Try reducing "hamming" from 3 or choose an alternative tag family.
$
```
The areas of code that were from some previous development/debugging that had exit or print statements have been enclosed in compiler removed `#if 0`/`#endif` sections although I guess they could just be commented out or simply removed.  I think this makes more sense then trying to wrap them into `asserts` given they currently aren't even active in the debug build but I can tray and change this if you still prefer the `assert.`

apriltag.c
```diff
@@ -295,8 +295,8 @@ static void quick_decode_codeword(apriltag_family_t *tf, uint64_t rcode,
 {
     struct quick_decode *qd = (struct quick_decode*) tf->impl;
 
-    // quick_decode might be NULL if adding detector_add_family_bits() failed
-    for (int ridx = 0; qd!=NULL && ridx < 4; ridx++) {
+    // quick_decode might be NULL if detector_add_family_bits() failed
+    for (int ridx = 0; qd!=NULL &&ridx < 4; ridx++) {
```
christian-rauch: All those "fixups" should be squashed into the original commit.

apriltag.c
```diff
@@ -296,20 +297,20 @@ static void quick_decode_codeword(apriltag_family_t *tf, uint64_t rcode,
...
-                }
+            if (qd->entries[bucket].rcode == rcode) {
+                *entry = qd->entries[bucket];
+                entry->rotation = ridx;
+                return;
             }
+        }
 
-            rcode = rotate90(rcode, tf->nbits);
+        rcode = rotate90(rcode, tf->nbits);
```
christian-rauch, line 314: This should all be part of your earlier commit "Tweak quick_decode check". Please "fixup" those commits.

apriltag.c
```diff
@@ -265,29 +266,29 @@ static void quick_decode_init(apriltag_family_t *family, int maxhamming)
...
+    //             }
+    //             run = 0;
+    //         } else {
+    //             run ++;
+    //             longest_run = imax(longest_run, run);
+    //         }
+    //     }
+
+    //     debug_print("quick decode: longest run: %d, average run %.3f\n", longest_run, 1.0 * run_sum / run_count);
+    // }
```
christian-rauch: Don't comment code. This should either be guarded by some compiler flag so it can be activated if required, or it should be deleted entirely.

common/debug_print.h
```diff
@@ -38,7 +38,7 @@ either expressed or implied, of the Regents of The University of Michigan.
 #define debug_print(fmt, ...) \
         do { if (DEBUG) fprintf(stderr, "%s:%d:%s(): " fmt, __FILE__, \
                                 __LINE__, __func__, __VA_ARGS__); fflush(stderr);} while (0)
-#endif#else
+#else
```
christian-rauch, line 43: Squash into original commit.

apriltag.c
```diff
@@ -291,16 +291,18 @@ static void quick_decode_init(apriltag_family_t *family, int maxhamming)
...
 {
     struct quick_decode *qd = (struct quick_decode*) tf->impl;
 
+    for (int ridx = 0; ridx < 4; ridx++) {
     // quick_decode might be NULL if detector_add_family_bits() failed
     for (int ridx = 0; qd!=NULL && ridx < 4; ridx++) {
 
         for (int bucket = rcode % qd->nentries;
+             qd->entries[bucket].rcode != UINT64_MAX;
+             bucket = (bucket + 1) % qd->nentries) {
```
christian-rauch, line 305: All those changes are reverting something that you did in an earlier commit "remove formatting changes" and it will be changed in the following "remove formatting changes". Squash this all together logically into the commits that original change this.

christian-rauch (changes requested):
Can you rebase your PR on top of master so that it appears that all commits have been created after each other and also squash all those fixup commits ("typo") into the original commits?

Currently, your PR contains multiple similar commits (e.g. "Replaced calls to exit()" and "Removed exit() calls from library") that seem to do the same and were resolved later. Also, you have some commits that fix your own commits (a.k.a. "fixup"). E.g. your commit "Tweak quick_decode check" is later changed by the two "typo" and "fixing indentation" commits. There are many more commits that change the code back and forth, e.g. there are commits where you comment-out code just to later guard this with a `#if 0`.

Especially having multiple commits that change the same code section multiple times makes it very hard to review this PR. When proposing a PR, your commit history should properly document which changes are made to the code base. Currently, this is very hard due to all these fixup commits.

apriltag.c
j-medland: I have tidied up all the commits into a single commit with (hopefully) clear and minimal changes on the most recent merge.

Hopefully this should make it easier to review.

christian-rauch (approved):

PR #236, opened 2022-06-16
Use official ros-tooling actions for CI

christian-rauch:
This is a proposal to use the `ros-tooling` actions [`setup-ros`](https://github.com/ros-tooling/setup-ros) and [`action-ros-ci`](https://github.com/ros-tooling/action-ros-ci) as those support Ubuntu, Windows and macOS, as well as custom CMake settings per package.

The Ubuntu workflow runs in the official ubuntu docker images in order to make sure that all dependencies are properly defined and resolved. The Windows and macOS pipeline runs on GitHub's custom configuration for the convenience of having all the build tools already installed.


wxmerkt (commented):
Please add the ROS-Industrial one back. It runs prerelease tests. Otherwise good to go

wxmerkt:
> I am curious about the pre-release tests. Is this checking if a bloom release is possible? Or is this running some unit tests?

It is replicating part of the buildfarm process. E.g. I can have tests passing in a normal workspace on my machine but fail in the prerelease test -- there is a strong correlation between those failures and when the buildfarm actually fails.

wxmerkt (approved):

PR #250, opened 2022-09-03
declare binary parameters as bool

christian-rauch:
Some of the `apriltag_detector_t` parameters are meant to be used as binary flags but stored as an integer with a much larger value range. This makes it difficult from an API perspective to directly grasp the meaning of a parameter. I am proposing here to change the type of `refine_edges`, `debug` and `qtp.deglitch` from `int` to `bool`.

While this is an API change, the impact should be low since previously used `int`s should be cast automatically for `bool`en values.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.1)
-project(apriltag VERSION 3.2.0 LANGUAGES C CXX)
+project(apriltag VERSION 3.3.0 LANGUAGES C CXX)
 
 if(POLICY CMP0077)
     cmake_policy(SET CMP0077 NEW)
--- a/apriltag.c
+++ b/apriltag.c
@@ -367,7 +367,7 @@ apriltag_detector_t *apriltag_detector_create()
 
     td->qtp.max_line_fit_mse = 10.0;
     td->qtp.cos_critical_rad = cos(10 * M_PI / 180);
-    td->qtp.deglitch = 0;
+    td->qtp.deglitch = false;
     td->qtp.min_white_black_diff = 5;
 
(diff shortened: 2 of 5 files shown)


christian-rauch:
@wxmerkt I bumped the version to 3.3 as we accumulated quite a few changes since the last release and the API changed slightly. After this has been merged, I (or you) would push a new tag to the repo.

Regarding the bloom releases, are those released automatically to `rolling` once the repo has been tagged or does someone from the release team have to trigger a new bloom release?

wxmerkt (approved):

wxmerkt:
> Regarding the bloom releases, are those released automatically to rolling once the repo has been tagged or does someone from the release team have to trigger a new bloom release?

We need to run releases manually - ideally to all distros. Beyond an API change, I suspect this may also introduce ABI changes which may cause issues with downstream packages unless rebuilt?

acshi (approved):

christian-rauch:
> > Regarding the bloom releases, are those released automatically to rolling once the repo has been tagged or does someone from the release team have to trigger a new bloom release?
> 
> We need to run releases manually - ideally to all distros. Beyond an API change, I suspect this may also introduce ABI changes which may cause issues with downstream packages unless rebuilt?

Yes, I also think there may be ABI changes. I am not sure how compilers represent `bool` values; it could be `int`, in which case this would still work. If in doubt, I would just re-release the downstream packages.

wxmerkt:
We cannot just re-release downstream packages since it's such a widely used package.

Also, I believe some of the motivation behind avoiding the use of `bool` might have been a desire to have compatibility with old C standards (perhaps for some embedded applications?). We might have to note the minimum requirement as such in the README/CMake settings if not set already.

PR #251, opened 2022-09-03
version the pkg-config file and use standard CMake for formatting

christian-rauch:
I noticed that the version on the `apriltag.pc` file was hard coded. CMake provides `configure_file` to format such a file automatically, including the version.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,11 +111,10 @@ export(TARGETS apriltag
     FILE ${generated_dir}/${targets_export_name}.cmake)
 
 
-# install pkgconfig related files
-FILE(READ apriltag.pc.in PKGC)
-STRING(REGEX REPLACE "^prefix=" "prefix=${CMAKE_INSTALL_PREFIX}" PKGC_CONF "${PKGC}" )
-FILE(WRITE ${PROJECT_BINARY_DIR}/apriltag.pc ${PKGC_CONF})
-install(FILES "${PROJECT_BINARY_DIR}/apriltag.pc" DESTINATION "${CMAKE_INSTALL_LIBDIR}/pkgconfig/")
+# install pkgconfig file
+configure_file(${PROJECT_NAME}.pc.in ${PROJECT_NAME}.pc @ONLY)
+install(FILES "${CMAKE_BINARY_DIR}/${PROJECT_NAME}.pc"
+        DESTINATION "${CMAKE_INSTALL_LIBDIR}/pkgconfig")
 
 
 # Python wrapper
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -1,5 +1,8 @@
 cmake_minimum_required(VERSION 3.1)
-project(apriltag VERSION 3.3.0 LANGUAGES C CXX)
+project(apriltag
+        VERSION 3.3.0
+        DESCRIPTION "AprilTag is a visual fiducial system popular for robotics research."
```
wxmerkt, line 4: This is failing for old CMake versions (here on the re-added 16.04 config). Does it make sense to remove the DESCRIPTION attribute?

```
CMake Error: Could not find cmake module file: CMakeDetermineDESCRIPTIONCompiler.cmake
  CMake Error: Error required internal CMake variable not set, cmake may be not be built correctly.
  Missing variable is:
  CMAKE_DESCRIPTION_COMPILER_ENV_VAR
  CMake Error: Error required internal CMake variable not set, cmake may be not be built correctly.
  Missing variable is:
  CMAKE_DESCRIPTION_COMPILER
  CMake Error: Could not find cmake module file: /root/target_ws/build/apriltag/CMakeFiles/3.5.1/CMakeDESCRIPTIONCompiler.cmake
  CMake Error: Could not find cmake module file: CMakeDetermineAprilTag is a visual fiducial system popular for robotics research.Compiler.cmake
  CMake Error: Error required internal CMake variable not set, cmake may be not be built correctly.
```

CMakeLists.txt
```diff
@@ -1,5 +1,8 @@
 cmake_minimum_required(VERSION 3.1)
-project(apriltag VERSION 3.3.0 LANGUAGES C CXX)
+project(apriltag
+        VERSION 3.3.0
```
wxmerkt, line 3: While we are it - does it make sense to read the version from the package.xml such that we don't maintain the numbering in two places?

CMakeLists.txt
christian-rauch: Sure, if the `DESCRIPTION` is the only thing that makes old CMake and Ubuntu fail, then we can remove it and actually just hard-code it in the .pc file.

CMakeLists.txt
christian-rauch: Is this possible? I know you can run custom commands in CMake, but I've not seen a project that parses the xml to extract the `version` tag. If this is possible with onboard tools (i.e. without installing additional software) then this would be an improvement to the error-prone manual version updates.

I've seen CMake examples getting the version from the git tag. But in any case, I don't think there is a way to set the `package.xml` version from either git or CMake.

wxmerkt (approved):

PR #252, opened 2022-09-04
Change WARNING to STATUS for building demo when OpenCV is not found

wxmerkt:


christian-rauch (approved):
Good point. I also got slightly annoyed by this error message showing up in colcon builds, although these demo programs have no relevance when `apriltag` is used as a library for another project.

Issue #256, opened 2022-09-09
Packaging suggestions


dkogan:
Hi. I'm the Debian maintainer. I just released the 3.3.0 release into Debian; it's there now. Ubuntu will pick it up when they make a new release, in a month probably. Thanks for maintaining this library. I have some notes:

- The top-level README in this project has a lengthy section on installing and building. I think this section should say "If you're running a recent Debian or Ubuntu, you should 'apt install apriltag'. Otherwise follow these complicated instructions". Also, if you're using the distro, stuff doesn't go into `/usr/local` or `/opt`; maybe the README doesn't need to talk about that at all.
- The Debian build has a few patches. Some are Debian-specific, but some you should consider merging. The patches live here: https://salsa.debian.org/science-team/apriltag/-/tree/master/debian/patches
  - https://salsa.debian.org/science-team/apriltag/-/blob/master/debian/patches/0002-I-explicitly-expose-ONLY-the-symbols-I-want-to-expor.patch: you should consider merging this. It explicitly marks which functions should be exported in the library. It works with the later patch 0006, which changes the default symbol visibility to "hidden". This allows you to explicitly define the functions that constitute your API. Otherwise all functions are your API, and you have more stuff to think about when making releases, and worrying about backwards compatibility
  - https://salsa.debian.org/science-team/apriltag/-/blob/master/debian/patches/0005-Libraries-refer-to-the-most-recent-version.patch: you should merge this, or better yet, make it automatically select the proper version
  - https://salsa.debian.org/science-team/apriltag/-/blob/master/debian/patches/0006-Split-the-code-into-two-libraries.patch: reworks the build to produce two separate shared objects. One has just the apriltag detector, the other has the utility functions. This separation is useful to not require the user to link in symbols with names like `getopt_create`. This could easily conflict with other symbols in a user's program, and there's no particular reason for the apriltag library to provide its own flavor of getopt to the world
  - https://salsa.debian.org/science-team/apriltag/-/blob/master/debian/patches/0007-added-simple_demo-uses-JUST-the-core-apriltag-functi.patch: is a demo apriltag program that links ONLY with the apriltag detector library, and processes images on the commandline. This is shipped in the packages as `/usr/bin/apriltag`. Merging this would be nice, I think

Thanks for maintaining this library!

christian-rauch:
- Patch 2 makes sense. But I think usually this is done with some macro that switches between OSes and compilers (e.g. support Windows).
[...]

PR #264, opened 2022-10-24
Build as sub components with defined exports

j-medland:
My first pass at a PR related to #256

The aim here is to provide sub-component libraries for users who don't need/want all the bundled utility functions.

I have maintained the behaviour of the existing `libapriltag.so` library for backwards compatibility but added the following libraries (with associated cmake exports)

`libapriltag-detector.so` _just the apriltag detector with the exports limited to the detector API_
`libapriltag-utils.so` _the utility functions used in the examples_
`libapriltag-tags.so` _all the tag libraries_
`libapriltag-<tag_family>.so` _each tag family as a standalone library for users who don't need every tag library_

I have added the two functions to the apriltag-detector API to compliment the existing `apriltag_to_image` which provides minimal functionality to write that `image` to a `pnm` file and then deallocate it (`apriltag_image_write_pnm`& `apriltag_image_destroy`)

Control of exports is achieved via macro definition and [cmake's `generate_export_headers`](https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html) although its usage is slightly complicated by the fact that multiple targets use the same `apriltag_export.h`. The [`WINDOWS_EXPORT_ALL`](https://cmake.org/cmake/help/v3.25/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html) target property (requires cmake 3.4) is used to export all functions from libraries on Windows. Having Windows defines means this library builds usable DLLs on Windows.

I have made a big change to the project's structure - separating the library components with their own `CMakeLists.txt` should hopefully make life easier for longer-term maintenance and moving away from globs and placing the API headers (those distributed with the binaries) into an `include/apriltag` should provide better control for users integrating this code into other projects.

Happy to hear your thoughts and discuss any changes.

John


j-medland:
Ah, looks like from the tests I forgot I was trying some newer cmake features. I think rolling back to 3.5 provides everything I need. 3.12 just provides some handy generator expressions.

j-medland:
I am currently tweaking this against the CI system so once that all passes, I will squash everything but feel free to take a look now if you want.

John

j-medland:
I think this is now all ready for review (after some rebase/squash commit mistakes on my part).

I am happy to hear feedback and make changes as folks see fit.

I did not integrate a command-line-interface application as @dkogan suggested in #256 but I believe these changes should make packaging as they suggest require fewer patches.

This PR also includes the optional builds of the examples as mentioned in #266.

As I mentioned in my comments, managing the exports also allows this library to be build as a DLLs. I have also had success creating a vcpkg portfile based on this commit which would allow easier integration into other projects and address microsoft/vcpkg#12171.

christian-rauch:
> Note - it looks `rosdep` is missing from the macOS build and causing it to fail

This must be something that changed recently. This pipeline used to work previously.

Can you try to update the `ros-tooling/setup-ros@v0.3` action to `ros-tooling/setup-ros@v0.4`? Maybe this fixes the issue.

christian-rauch:
Looking at the CI logs, `rosdep` actually gets installed by pip but the command cannot be found. I believe this is a bug in the GitHub action [`ros-tooling/setup-ros`](https://github.com/ros-tooling/setup-ros) that is caused by updating the macOS version on the runners (https://github.blog/changelog/2022-10-03-github-actions-jobs-running-on-macos-latest-are-now-running-on-macos-12/).

Since this is a fairly huge PR, I would now wait for the reviews anyway. In the meantime, the issue in the action maybe gets fixed. Otherwise, we can just flag the macOS CI as optional or remove it again.

christian-rauch:
Can you rebase your PR to fix the CI issue?

CMakeLists.txt
```diff
@@ -38,150 +44,20 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
...
-
-
-# install pkgconfig file
-configure_file(${PROJECT_NAME}.pc.in ${PROJECT_NAME}.pc @ONLY)
-install(FILES "${CMAKE_BINARY_DIR}/${PROJECT_NAME}.pc"
-        DESTINATION "${CMAKE_INSTALL_LIBDIR}/pkgconfig")
-
-
-# Python wrapper
+## Python wrapper
```
christian-rauch, line 50: This should be a single `#` just like in the other cases.

CMakeLists.txt
```diff
@@ -38,150 +44,20 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
...
-else()
-    message(STATUS "OpenCV not found: Not building demo")
-endif(OpenCV_FOUND)
-
-# install example programs
-install(TARGETS apriltag_demo RUNTIME DESTINATION bin)
+if(BUILD_EXAMPLES)
+    # build examples
+    add_subdirectory(example)
+endif()
```
christian-rauch, line 63: There is an end-of-line missing here.

libs/apriltag/CMake/apriltagHelperFunctions.cmake
```diff
@@ -0,0 +1,28 @@
...
+function(set_apriltag_export_all target_name)
+    set_target_properties(${target_name} PROPERTIES 
+        C_VISIBILITY_PRESET default 
+        CXX_VISIBILITY_PRESET default
+        VISIBILITY_INLINES_HIDDEN OFF
+        WINDOWS_EXPORT_ALL_SYMBOLS ON
+        # multiple libs sharing same auto-generated header so override default to use apriltag_EXPORTS
+        DEFINE_SYMBOL apriltag_EXPORTS
+    )
+endfunction()
```
christian-rauch, line 28: new line missing

libs/apriltag/CMake/apriltagHelperFunctions.cmake
```diff
@@ -0,0 +1,28 @@
...
+        C_VISIBILITY_PRESET hidden 
+        CXX_VISIBILITY_PRESET hidden
+        VISIBILITY_INLINES_HIDDEN ON
+        WINDOWS_EXPORT_ALL_SYMBOLS OFF
+        # multiple libs sharing same auto-generated header so override default to use apriltag_EXPORTS
+        DEFINE_SYMBOL apriltag_EXPORTS
+    )
+endfunction()
+
+function(set_apriltag_export_all target_name)
```
christian-rauch, line 19: Can you add a 1~2 line comment to each of the three functions to explain what they are doing?

libs/apriltag/CMakeLists.txt
```diff
@@ -0,0 +1,298 @@
+include(GNUInstallDirs)
+include(GenerateExportHeader)
+include(CMake/apriltagHelperFunctions.cmake)
+
+### Build Apriltag Libraries
+## The following targets are built
+##      apriltag-detector (Apriltag's core functionality with only API functions exported)
+##      apriltag-tags (Apriltag's tag libraries)
+##      apriltag-<tag_family_name> e.g apriltag-tagStandard41h12 (Individual libraries for each tag family)
```
christian-rauch, line 9: Why is there a library containing all tags and individual libraries per tag? Until we can load tags via plugins, I would suggest that one library containing all standard tags is sufficient.

libs/apriltag/CMakeLists.txt
```diff
@@ -0,0 +1,298 @@
...
+    set_target_properties(${TAG_TARGET} PROPERTIES EXPORT_NAME ${TAG_FAMILY})
+    target_link_libraries(${TAG_TARGET} PUBLIC ${PROJECT_NAME}-include)
+
+    LIST(APPEND TARGETS_TAGS ${TAG_TARGET})
+    LIST(APPEND HEADERS_TAGS ${TAG_FAMILY_HEADER})
+    LIST(APPEND SRC_TAGS ${TAG_FAMILY_SRC})
+endforeach()
+
+# All-Tag-Families library
+add_library(${PROJECT_NAME}-tags ${SRC_TAGS} ${SRC_HEADERS})
```
christian-rauch, line 166: I don't see `SRC_HEADERS` defined anywhere. I think this is empty and can be removed.

christian-rauch (changes requested):
This is great work. I like the ability to only link the core functionality. Also, the new file layout makes it easier to understand the different parts of the repo.

The "apriltag" folder now contains a single `CMakeLists.txt` defining multiple libraries. To better reflect this split, it would also make sense to split the "apriltag" folder into a file layout that mirrors those libraries with individual source folders and `CMakeLists.txt` per library. E.g. there could be `detector` and `tags` folder containing the sources and `CMakeLists.txt` for the `apriltag-detector` and `apriltag-tags`.

Some sources in the `common` folder, such as `homography.c` and `workerpool.h`, are only used in the `detector` library. There might be more `common` sources that are only used in one of the libraries. I think it would make sense to move source that only belong to one of the libraries out of `common` into the folder for that specific library (`detector`, `tags`, ...).

Finally, the `apriltag.pc` references only the now legacy `apriltag` library. Can CMake generate new pkg-config files for the libraries (I know meson can do this)? Otherwise, this new improved separation of functionality will not be available outside of CMake.

j-medland:
Thanks for the feedback. I'll hopefully have some time over the next couple of weeks to work through the changes you have proposed.

I'll update when things are ready for review again.

Cheers

John

christian-rauch (changes requested):
What is the status of this PR? Is there still interest in merging this?

j-medland:
I am still interested and hoping to give it another pass with recent updates and your suggestions soon.

Cheers!

eli-schwartz:
> Finally, the `apriltag.pc` references only the now legacy `apriltag` library. Can CMake generate new pkg-config files for the libraries (I know meson can do this)? Otherwise, this new improved separation of functionality will not be available outside of CMake.

cmake doesn't have a native mechanism to do this, but you can just keep track of library names for dependencies in a second list, and format the pkg-config file as a template.

Issue #90, opened 2020-06-21
Obtaining euler angles from apriltag_pose_t?


FROGbots-4634:
Once I have obtained an `apriltag_pose_t`, how can I get the euler angles from the rotation matrix provided at `pose.R`?

mkrogius:
You can use the method common/doubles_mat_to_quat to get a quaternion from the matrix. Then you can use common/doubles_quat_to_rpy to convert that quaternion to RPY.

vladLozkin:
Hi, mkrogius,
Im looking for obtaining euler angles from post also but I cannot find the functions that you mentioned in the code...

can you point to a more specific location / show how to call them from cpp?

vladLozkin:
> You can use the method common/doubles_mat_to_quat to get a quaternion from the matrix. Then you can use common/doubles_quat_to_rpy to convert that quaternion to RPY.

Hi, mkrogius,
Im looking for obtaining euler angles from post also but I cannot find the functions that you mentioned in the code...

can you point to a more specific location / show how to call them from cpp?

mkrogius:
The methods are declared in this header file: https://github.com/AprilRobotics/apriltag/blob/master/common/doubles.h

which templates the code from this header file: https://github.com/AprilRobotics/apriltag/blob/master/common/doubles_floats_impl.h#L456

chobitsfan:
doubles_mat_to_quat assumes a 4x4 matrix, but pose.R is a 3x3 matrix. You need to re-arrange pose.R before give it to doubles_mat_to_quat

PR #265, opened 2022-10-25
add 3x3 rotation matrix to quaternion function

chobitsfan:
In https://github.com/AprilRobotics/apriltag/issues/90#issuecomment-656802651, @mkrogius suggested using `doubles_mat_to_quat` to get a quaternion from the matrix. However, rotation matrix in `apriltag_pose_t` is a 3x3 matrix and `doubles_mat_to_quat` expects a 4x4 matrix. It would be easier to use if we have a `doubles_mat33_to_quat`.

### DIFF
--- a/common/doubles_floats_impl.h
+++ b/common/doubles_floats_impl.h
@@ -453,6 +453,40 @@ static inline void TFN(s_mat_to_xyz)(const TNAME M[16], TNAME xyz[3])
     xyz[2] = M[11];
 }
 
+static inline void TFN(s_mat33_to_quat)(const TNAME M[9], TNAME q[4])
+{
+    double T = M[0] + M[4] + M[8] + 1.0;
+    double S;
+
+    if (T > 0.0000001) {
+        S = sqrt(T) * 2;
+        q[0] = (TNAME)(0.25 * S);
+        q[1] = (TNAME)((M[7] - M[5]) / S);
+        q[2] = (TNAME)((M[2] - M[6]) / S);
+        q[3] = (TNAME)((M[3] - M[1]) / S);
+    } else if (M[0] > M[4] && M[0] > M[8]) {   // Column 0:
+        S = sqrt(1.0 + M[0] - M[4] - M[8]) * 2;
+        q[0] = (TNAME)((M[7] - M[5]) / S);
+        q[1] = (TNAME)(0.25 * S);
+        q[2] = (TNAME)((M[3] + M[1]) / S);
+        q[3] = (TNAME)((M[2] + M[6]) / S);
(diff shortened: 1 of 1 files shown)


christian-rauch:
Thank you for the patch. Where is this new function used? I cannot find references to `doubles_mat_to_quat`. Can you provide more details on why this new function is required?

chobitsfan:
Hi @christian-rauch Thank you for reviewing

> I cannot find references to doubles_mat_to_quat

Yes. `doubles_mat_to_quat` is defined but never used. But in https://github.com/AprilRobotics/apriltag/issues/90, @mkrogius suggests it can be used (with `doubles_quat_to_rpy`) to obtain euler angles from apriltag_pose_t. I found it is quite useful.  However, rotation matrix in apriltag_pose_t is a 3x3 matrix and doubles_mat_to_quat expects a 4x4 matrix.

common/doubles_floats_impl.h
```diff
@@ -453,6 +453,40 @@ static inline void TFN(s_mat_to_xyz)(const TNAME M[16], TNAME xyz[3])
     xyz[2] = M[11];
 }
 
+static inline void TFN(s_mat33_to_quat)(const TNAME M[9], TNAME q[4])
```
christian-rauch, line 456: Maybe call this function "rot_mat_to_quat"? "mat33" sounds as if it accepts any 3x3 matrix. Do you think we need to check that the matrix provided is actually a rotation matrix?

common/doubles_floats_impl.h
```diff
@@ -453,6 +453,40 @@ static inline void TFN(s_mat_to_xyz)(const TNAME M[16], TNAME xyz[3])
...
+        q[1] = (TNAME)((M[3] + M[1]) / S);
+        q[2] = (TNAME)(0.25 * S);
+        q[3] = (TNAME)((M[7] + M[5]) / S);
+    } else {                                    // Column 2:
+        S = sqrt(1.0 + M[8] - M[0] - M[4]);
+        q[0] = (TNAME)((M[3] - M[1]) / S);
+        q[1] = (TNAME)((M[2] + M[6]) / S);
+        q[2] = (TNAME)((M[7] + M[5]) / S);
+        q[3] = (TNAME)(0.25 * S);
+    }
```
christian-rauch, line 485: Let's try to avoid code duplication with `mat_to_quat` here. Either have `mat_to_quat` reuse this very function with a submatrix (the best option IMHO) or extend the 3x3 rotation to 4x4 transformation matrix and reuse `mat_to_quat` here.

christian-rauch (changes requested):

PR #266, opened 2022-11-02
Added a cmake option for building example executables

jacobhuesman:
I have a customized OpenCV install that caused the opencv example build to fail. I've added an cmake option to allow users to optionally disable building the examples.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,7 @@ if(POLICY CMP0077)
     cmake_policy(SET CMP0077 NEW)
 endif()
 option(BUILD_SHARED_LIBS "Build shared libraries" ON)
+option(BUILD_EXAMPLES "Build example executables" ON)
 option(ASAN "Use AddressSanitizer for debug builds to detect memory issues" OFF)
 
 if (ASAN)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -160,28 +161,30 @@ install(FILES ${PROJECT_BINARY_DIR}/apriltag${PY_EXT_SUFFIX} DESTINATION ${PY_DE
...
+        target_link_libraries(opencv_demo apriltag ${OpenCV_LIBRARIES})
+        set_target_properties(opencv_demo PROPERTIES CXX_STANDARD 11)
+        install(TARGETS opencv_demo RUNTIME DESTINATION bin)
+    else()
+        message(STATUS "OpenCV not found: Not building demo")
+    endif(OpenCV_FOUND)
+
+    # install example programs
+    install(TARGETS apriltag_demo RUNTIME DESTINATION bin)
+endif()
```
christian-rauch, line 190: This is missing a new line at the end of the file.

christian-rauch:
Can you give details why the compilation fails? Maybe we can just use a required version or components to check if a "normal" OpenCV is installed.

CMakeLists.txt
jacobhuesman: Fixed

jacobhuesman:
> Can you give details why the compilation fails? Maybe we can just use a required version or components to check if a "normal" OpenCV is installed.

There's an additional Eigen dependency that get's pulled in. 

I'm only interested in the core library since I have existing code that uses AprilTags, so being able to explicitly skip building the examples is useful for me. 

The BUILD_EXAMPLES option is based on how OpenCV enables and disables building their examples: https://docs.opencv.org/4.x/db/d05/tutorial_config_reference.html

christian-rauch:
Can you rebase your PR to fix the CI issue?

jacobhuesman:
Rebased the PR

PR #274, opened 2023-04-14
Change rpath of the cython file

powei-lin:
Change the runtime path of cythonized file to `$ORIGIN`. So user can copy `libapriltag.so` and `apriltag.cpython-310-x86_64-linux-gnu.so` to other places and still be able to use them in python.


CMakeLists.txt
```diff
@@ -149,7 +149,8 @@ if (NOT Python3_NOT_FOUND AND NOT Numpy_NOT_FOUND AND PYTHONLIBS_FOUND AND BUILD
         COMMAND ${CMAKE_C_COMPILER} ${PY_CFLAGS} -I${PROJECT_BINARY_DIR} -c -o apriltag_pywrap.o ${CMAKE_CURRENT_SOURCE_DIR}/apriltag_pywrap.c
         DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/apriltag_pywrap.c ${PROJECT_BINARY_DIR}/apriltag_detect.docstring.h ${PROJECT_BINARY_DIR}/apriltag_py_type.docstring.h)
     add_custom_command(OUTPUT apriltag${PY_EXT_SUFFIX}
-        COMMAND ${PY_LINKER} ${PY_LDFLAGS} -Wl,-rpath,lib apriltag_pywrap.o $<TARGET_FILE:apriltag> -o apriltag${PY_EXT_SUFFIX}
+        COMMAND ${PY_LINKER} ${PY_LDFLAGS} -Wl,-rpath=tmppath apriltag_pywrap.o $<TARGET_FILE:apriltag> -o apriltag${PY_EXT_SUFFIX}
+        COMMAND sed -i 's/tmppath/$$ORIGIN/' apriltag${PY_EXT_SUFFIX}
```
christian-rauch, line 153: Is it not possible to directly set `-Wl,-rpath=$$ORIGIN`?

powei-lin: It tried, but it seems like it's impossible to do it at compile time

powei-lin: You can use `objdump -x apriltag.cpython-38-x86_64-linux-gnu.so | grep RUNPATH` after compile. I don't think it's possible to set it to `$ORIGIN` during compile.

christian-rauch: Would it work as proposed in the [Ubuntu snap guide](https://snapcraft.io/blog/the-new-classic-confinement-in-snaps-even-the-classics-need-a-change) (using linker flags `-Wl,-rpath=\$ORIGIN/lib`)?

powei-lin: The problem is not `gcc`.  It's cmake. Every `$` is being replace in line 152.

christian-rauch: Did you look into [`BUILD_RPATH_USE_ORIGIN`](https://cmake.org/cmake/help/latest/prop_tgt/BUILD_RPATH_USE_ORIGIN.html)? That looks exactly like what you are looking for. I also found some information here: https://stackoverflow.com/questions/58360502/how-to-set-rpath-origin-in-cmake.

powei-lin: I found out that `add_custom_command` is not able to do that. So I changed it to `add_custom_target` and then I found out that `add_custom_target` can easily reserve `$ORIGIN` by adding `VERBATIM`

christian-rauch:
Since you are changing the same section twice, can you just squash your two commits into one?

powei-lin:
@christian-rauch Done

PR #276, opened 2023-06-05
Duplicate definitions

v1otusc:
Duplicate definitions of zhash->nentries

### DIFF
--- a/common/zhash.c
+++ b/common/zhash.c
@@ -84,7 +84,6 @@ zhash_t *zhash_create_capacity(size_t keysz, size_t valuesz,
     zh->entrysz = 1 + zh->keysz + zh->valuesz;
 
     zh->entries = calloc(zh->nentries, zh->entrysz);
-    zh->nentries = nentries;
 
     return zh;
 }


christian-rauch:
Can you explain where this is duplicated? `zhash_t *zh` is initialised just before this. Also, the commit message is not very explanatory.

christian-rauch:
> It's duplicated in L82 just above the fold, so this is correct.

Right. I opened the file on GitHub with this change already applied.

> Unclear whether to remove L87 as proposed or alternatively L82 since the allocation related to nentries happens in L86 so it would be 'closer together'.

We should remove L87 as suggested. Removing L82 would assume 0 for `zh->nentries` in the allocation at L86.

common/zhash.c
```diff
@@ -84,7 +84,6 @@ zhash_t *zhash_create_capacity(size_t keysz, size_t valuesz,
     zh->entrysz = 1 + zh->keysz + zh->valuesz;
 
     zh->entries = calloc(zh->nentries, zh->entrysz);
-    zh->nentries = nentries;
```
christian-rauch, line 87: This needs a better commit message than "Duplicate definitions" as this could be related to any part of the code.

christian-rauch (changes requested):

Issue #278, opened 2023-06-08
Improve documentation concerning estimated pose


Lecrapouille:
This is not an issue but a hybrid pull request. I cannot make a pull request because this both impact comments in the code and changes inside the wiki.

### Update a comment in the code

This structure has no comments. 

https://github.com/AprilRobotics/apriltag/blob/df4aee5b511d2176e82bf9585469c5543e741486/apriltag_pose.h#L23

Here what I suggest adding:

```
/**
 * This struct holds the transformation from the camera optical frame to
 * the April tag frame. The pose refers to the position of the tag within
 * the camera frame.
 *
 * If you want to get the camera pose with respect to the April tag frame
 * you have to apply the inverse of this transformation.
 */
typedef struct {
    matd_t* R; // Rotation matrix 3x3 of doubles.
    matd_t* t; // Translation matrix 3x1 of doubles.
} apriltag_pose_t;
```

### Update the wiki

I give the md file that includes 2 figures of coordinates

![coord_camera](https://github.com/AprilRobotics/apriltag/assets/17874458/3350fad2-f396-43a4-8de1-d0e3fa922952)
![coord_tag](https://github.com/AprilRobotics/apriltag/assets/17874458/53d4ea66-4d2f-46b2-b985-977dc8bed877)
[AprilTag-User-Guide.md](https://github.com/AprilRobotics/apriltag/files/11683568/AprilTag-User-Guide.md)

Figures are PNG but made with Draw.io so the "drawio code" is embedded in the PNG so you can open it with drawio to edit it.
The markdown file:
- I linked an issue for giving an example of OpenCv's Pnp solver
- I added a section concerning apriltag_pose_t and camera pose. @christian-rauch  @mkrogius please review it since I'm little rusty with transfer matrices.

christian-rauch:
You can send a PR which adds the comments to the `apriltag_pose_t` struct anyway. Not sure about the wiki. Someone else has to add those changes as you cannot send a PR to the wiki git (https://github.com/AprilRobotics/apriltag.wiki.git).

[...]

PR #279, opened 2023-06-12
Document the meaning of apriltag_pose_t struct

Lecrapouille:
Refer to https://github.com/AprilRobotics/apriltag/issues/278

### DIFF
--- a/apriltag_pose.h
+++ b/apriltag_pose.h
@@ -17,9 +17,14 @@ typedef struct {
     double cy; // In pixels.
 } apriltag_detection_info_t;
 
+/**
+ * This struct holds the transformation from the camera optical frame to
+ * the April tag frame. The pose refers to the position of the tag within
+ * the camera frame.
+ */
 typedef struct {
-    matd_t* R;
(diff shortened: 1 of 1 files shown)


apriltag_pose.h
```diff
@@ -17,9 +17,17 @@ typedef struct {
     double cy; // In pixels.
 } apriltag_detection_info_t;
 
+/**
+ * This struct holds the transformation from the camera optical frame to
+ * the April tag frame. The pose refers to the position of the tag within
+ * the camera frame.
+ *
+ * If you want to get the camera pose with respect to the April tag frame
+ * you have to apply the inverse of this transformation.
```
christian-rauch, line 26: I would say that this is redundant, i.e. this is common knowledge and applies to any transformation.

christian-rauch (changes requested):
I find "Comment to" too generic. Say that this documents the meaning of the `apriltag_pose_t` or similar.

Lecrapouille:
Ok I git pushed force

PR #280, opened 2023-06-15
Speed up cluster merge by batch copying arrays

bouk:
I was looking through the code and found that a lot of time was being spent on 'make clusters' so I did a quick look through why that is the case. I found the merge_clusters function to be inefficient, so I optimized it to do batch memcopies wherever possible.

This gives me a ~20% reduction in time spent for a large image.

## Before

Screenshot 2023-06-15 at 14 33 47

## After

Screenshot 2023-06-15 at 14 33 56


bouk:
@christian-rauch addressed your comments!

PR #281, opened 2023-06-20
Add wiki inside the project

Lecrapouille:
Since pull request on the wiki is difficult, better to move it inside the repo.


christian-rauch (changes requested):
Since the wiki is essentially a single markdown document, I would simply move that content to the [`Usage` section of the `README.md`](https://github.com/AprilRobotics/apriltag#usage). Since it's quite big, I would move this after `Install` and add quick links to the sections at the top.

Lecrapouille:
@christian-rauch updated I hope to have understood you well

christian-rauch:
I meant that the whole content of the single markdown file you added (`AprilTag-User-Guide.md`) should simply be added to the `README.md` section `Usage`, and that that section should move after `Install`.

The image file can be placed in the root of the repo as there are no other files for the documentation. You can place it inside a `doc` folder etc. if you plan to add additional images to the documentation.

You can keep the link as they are now but I would still recommend merging all the markdown documentation into the `README.md`.

christian-rauch (approved):

christian-rauch:
@mkrogius Are you ok with these changes? Once this is merged, can you remove the wiki from this repo in order to avoid confusion?

Lecrapouille:
@christian-rauch I've mde changes yes ! For removing the wiki I'm not sure I'm allowed to do this.

mkrogius:
I'm on board. Merging and deleting the original wiki

PR #286, opened 2023-07-20
Changed type to size_t for consistent type comparison

karina-ranadive:
In getopt.c, the variable pos was of type int and this comparison to the len variable, which is of type size_t, presents a security vulnerability for my team's utilization of this repository as pos is of a different type than len. I changed the pos variable to be of type size_t to enable pos and len to be consistently compared in the for loop where "pos < len". Similarly, in string_util.c, pos and haystack_len are compared as inconsistent types so to address this security vulnerability, I changed pos to be of type size_t. In pam.c, I changed idx to be of type size_t to address the same security vulnerability as for the files above.


christian-rauch:
Can you squash your commits together?

karina-ranadive:
Just squashed the commits, thanks!

PR #294, opened 2023-10-26
Optimize unionfind

bouk:
I figured out that a big part of the union find time spent was initializing the data array. I've optimized it by using two `memset`s instead of an array, which saves ~4% for an image sized 4896×3264

## Before

Screenshot 2023-10-26 at 13 42 26

## After

Screenshot 2023-10-26 at 13 42 13

### DIFF
--- a/common/unionfind.h
+++ b/common/unionfind.h
@@ -27,6 +27,7 @@ either expressed or implied, of the Regents of The University of Michigan.
 
 #pragma once
 
+#include <string.h>
 #include <stdint.h>
 #include <stdlib.h>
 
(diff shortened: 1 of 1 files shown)


christian-rauch:
Is this essentially saving you 5ms on the 4896×3264 image? That's a nice improvement.

Could you explain in the commit message a bit what the change is doing and why it improves the performance? Glancing over this, Isee it's moving the `parent` and `size` from the `struct ufrec` into the `struct unionfind`, turning them into a point for arrays and then initialising all elements in the list at once instead of looping over the array of `struct ufrec`?

bouk:
Expanded the commit message a bit. The savings is indeed from using `memset` instead of a for loop

christian-rauch:
Can you remove the concrete ~4% from the commit message again? I guess this value depends a lot on the image resolution and is not the same improvement in all cases.

bouk:
Sure, removed

christian-rauch (approved):

PR #297, opened 2023-10-28
fix compilation with WIN32_LEAN_AND_MEAN

### LINKED ISSUE #296: compatibility with WIN32_LEAN_AND_MEAN
This library is not compatible with `WIN32_LEAN_AND_MEAN` definition. It fails to find struct `timeval` which is normally found by `windows.h` header. But with `WIN32_LEAN_AND_MEAN` this header is excluded. 

We could directly include `winsock.h` header as it is specified in official documentation (https://learn.microsoft.com/en-us/windows/win32/api/winsock/ns-winsock-timeval).

https://github.com/AprilRobotics/apriltag/blob/b7bd75ca45d62ac431f10b78b69ea86198cbad14/common/time_util.h#L41

I can prepare a PR for you with this change.

mitjap:
This PR resolves #296. I changed `windows.h` header for `Winsock2.h`. This way the library still works even when used with `WIN32_LEAN_AND_MEAN` preprocessor definition.

I think even better solution would be to move function `gettimeofday` in `time_util.c` which is the only place where it is used. This way we can also move any required headers (either `windows.h`, `winsock.h` or `Winsock2.h`) from header file which would prevent any side effects to other libraries.

https://github.com/AprilRobotics/apriltag/blob/cc655ad6e3dadb19658276f7476dad5e92f05144/common/time_util.h#L41

### DIFF
--- a/common/time_util.h
+++ b/common/time_util.h
@@ -32,7 +32,7 @@ either expressed or implied, of the Regents of The University of Michigan.
 #include <time.h>
 
 #ifdef _WIN32
-#include <windows.h>
+#include <Winsock2.h>
 typedef long long suseconds_t;
 #endif
 


christian-rauch:
I added a new standard CMake pipeline to the CI that tests multiple compilers on different systems. Could you rebase your PR and add a setting for "WIN32_LEAN_AND_MEAN"?

christian-rauch:
I added `WIN32_LEAN_AND_MEAN` by default (https://github.com/AprilRobotics/apriltag/pull/372), verified that the Windows build fails with `error: declaration of 'struct timeval' will not be visible outside of this function [-Werror,-Wvisibility]`, and then verified that your `Winsock2.h` change fixes the issue again.

I am going to merge your PR and then rebase mine to make sure that `WIN32_LEAN_AND_MEAN` continues to work.

Issue #267, opened 2022-11-28
Build/install issue with shared libraries


parthapn98:
Hi Im building april tag using the commands under install here https://github.com/AprilRobotics/apriltag

if i build for -DBUILD_SHARED_LIBS=OFF
The install works properly, however i need the python wrapper for my work.

However when building with -DBUILD_SHARED_LIBS=ON
followed by 
cmake --build build --target install --clean-first
I get an error:
LINK : fatal error LNK1104: cannot open file 'Debug\apriltagd.lib'

and the file itself is missing in the directory.

Any suggestions?
Windows 10, python 3.9.13

christian-rauch:
The python wrapper (`BUILD_PYTHON_WRAPPER`) can only be used with shared libraries (see `cmake_dependent_option(BUILD_PYTHON_WRAPPER "Builds Python wrapper" ON BUILD_SHARED_LIBS OFF)`).

At least on Linux, the wrapper is built automatically if `BUILD_SHARED_LIBS` is `ON`. I haven't tried this on Windows. The Python wrapper may just not work on Windows. I can review a PR if someone implements this with a CI pipeline.

parthapn98:
is the build tested on unix/ redhat?
it worked fine on Ubuntu

however in windows the build+install without the shared libraries work
but with shared libraries they don't. can you recommend a way for me to share the data with you if you are willing to look into this?

christian-rauch:
> is the build tested on unix/ redhat? it worked fine on Ubuntu

I only ever use this on Ubuntu. I assume that this should work on any recent Linux distribution that provides the dependencies. I am not sure if you can compile this on any UNIX variant

> however in windows the build+install without the shared libraries work but with shared libraries they don't. can you recommend a way for me to share the data with you if you are willing to look into this?

Sorry, I don't use Windows and I also do not want to invest the time to set up a Windows development environment in a VM.

The Windows support was generously provided (amongst others) by @ryancwicks (https://github.com/AprilRobotics/apriltag/pull/234) and @VincentMarnier (https://github.com/AprilRobotics/apriltag/pull/240). They may be able to help you.

VincentMarnier:
Hello,
I think I might misunderstand something here as I successfuly can build for windows (with and without cross compiling) with both DBUILD_SHARED_LIBS ON and OFF.
I am using the "Developer Command Prompt for VS 2022" (with cmake installed by "Visual Studio Installer").
I run the following commands:
```
[...]

PR #311, opened 2024-01-22
enable Python bindings for Windows and macOS

christian-rauch:
The `FindPythonLibs` module has been deprecated since CMake 3.12. The newer `FindPython` provides a better and more portable way to build Python modules without the need for a dedicated script to determine the system configuration. This allows us now to build the Python bindings on Windows and macOS. This PR also changes the install path for the resulting Python module from the "user site" to something determined by the given CMake install path. This allows us to properly install the Python bindings when distributing binary packages.

Addresses https://github.com/AprilRobotics/apriltag/issues/267#issuecomment-1335871722.


christian-rauch:
@mkrogius @wxmerkt In order to make the Python bindings usable for binary distributed packages and other OSes, the "breaking change" in this PR is the install location for the Python module. I assume you are ok with this, as the new install location is the proper path determined from the CMake install path.

mkrogius:
For sure!

Issue #100, opened 2020-08-14
"No rule to make target '../libapriltag.a' " when compiling demo codes


WaterFriend:
Hi Everybody:
I am using the Ubuntu18.04 virtual machine (virtual box) as my environment. I have successfully installed OpenCV 3.4.8 and Apriltag library. But when I try to cmake the demo by following the official instructions "_make opencv_demo_" and "_make apriltag_demo_", the following errors show up:

```
$ cd example/
$ make opencv_demo 
make: *** No rule to make target '../libapriltag.a', needed by 'opencv_demo'.  Stop.
$ make apriltag_demo 
make: *** No rule to make target '../libapriltag.a', needed by 'apriltag_demo'.  Stop.
```

I try to ignore the error and directly run the demos, the following errors show up:

```
$ ./apriltag_demo 
apriltag_demo: error while loading shared libraries: libapriltag.so.3: cannot open shared object file: No such file or directory
$ ./opencv_demo
opencv_demo: error while loading shared libraries: libapriltag.so.3: cannot open shared object file: No such file or directory
```

But in fact, I have libapriltag.so.3 installed in /usr/local/lib as Apriltag library is successfully installed. Is anyone has any idea with these issues? Thanks for any advice

jeremi:
@WaterFriend I had the same issue, running `sudo ldconfig` resolved it.

Reihold-Chen:
I meet the same mistake, did you have solved it ?

Zhang-Qinghui:
I meet the same mistake, did you have solved it ?

christian-rauch:
@Reihold-Chen @Zhang-Qinghui Did running `sudo ldconfig` work for you?

Are you using `cmake` or `make`? The documentation has instructions for both. You can also try to run CMake with Ninja instead:
```
cmake . -GNinja
ninja
```

Hola39e:
I met the same mistake and got a solution. 
the `../libapriltag.a` file doesnt exsit in the right path when using `cmake` to build apriltag library, the lib was directly added to `/usr/local/lib`.
so change the `Makefile` file in /example,  delete `../libapriltag.a` text and add `pkg-config --libs apriltag` instead, then is could be compiled successfully.

```
apriltag_demo: apriltag_demo.o  
      @echo "   [$@]"
      @$(CC) -o $@ $^ $(LDFLAGS) `pkg-config --libs apriltag`
 
opencv_demo: opencv_demo.o
     @echo "   [$@]"
[...]

PR #313, opened 2024-01-24
remove old UNIX-specific Makefile

christian-rauch:
I would like to remove the Makefile since:
1. it is not tested on the CI
2. does not compile the Python bindings
3. confuses users how to build and install the apriltag library (https://github.com/AprilRobotics/apriltag/issues/100)

Also see the discussion here: https://github.com/AprilRobotics/apriltag/issues/210.


wxmerkt (commented):
I am okay with this since it's out of sync with the CMake and leads to different behaviour. We may want to highlight explicitly in CHANGELOG and in the version bump that it's been removed in case some downstream workflow dependent on it without explicit version tagging (e.g. off master)

PR #314, opened 2024-01-31
add tests

christian-rauch:
This adds tests on example images with a defined set of "ground truth" detections. I simply recorded the output of the detector once and compare to this in the tests. If the detections change in future, we can detect this with the tests.

There is some non-determinism in the order of the `zarray_t` of `apriltag_detection` returned by `apriltag_detector_detect`. I could resolve this across OSes, i.e. Linux, Windows and macOS will return the same order, but the order still differs across versions of the library for the same set of detections. Visually and by matching the outputs, one can see that they stayed the same. We may have to properly search between detected and true detections in the txt file to find mismatches.

| ver | 33369213973_9d9bb4cc96_c  | 34085369442_304b6bafd9_c | 34139872896_defdb2f8d9_c |
| ----| ------------- | ------------- | ------------- |
| 3.2 | debug_output | debug_output | debug_output |
| 3.3 | debug_output | debug_output | debug_output |
| 3.4 | debug_output | debug_output | debug_output |


jrepp:
really nice!

PR #316, opened 2024-02-04
bump minor version to prepare release

christian-rauch:
The last release was more than 1 year ago and many performance and compilation improvements have been committed since then. Let's create a new release for this.

The most notable breaking changes are the removal of the Makefile, some changes in the Python bindings installation path and the removal of unused parameters in some public functions which will break API if they are used directly.

I intend to use this release to create new binary package releases for the ROS repo and conda-forge.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required(VERSION 3.16)
-project(apriltag VERSION 3.3.0 LANGUAGES C)
+project(apriltag VERSION 3.4.0 LANGUAGES C)
 
 if(POLICY CMP0077)
     cmake_policy(SET CMP0077 NEW)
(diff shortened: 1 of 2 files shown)


wxmerkt (commented):
Thank you for preparing this. Can you please update the CHANGELOG. Did you check that the Python install path change doesn't break on ROS1 if it's used there?

christian-rauch:
Do you mean a specific CHANGELOG file? I couldn't find such a file in the repo. I would add those notes to the release notes on the GitHub release page.

Regarding the Python bindings, they were never built and installed for the ROS package. Only since https://github.com/AprilRobotics/apriltag/pull/311 are we specifying the Python dependencies for the ROS package. The ROS package that will be released for 3.4 will install those Python bindings to `/opt/ros/humble/lib/python3.10/site-packages/apriltag.cpython-310-x86_64-linux-gnu.so`. I tested this with the [ros-deb-builder-action](https://github.com/jspricke/ros-deb-builder-action).

Issue #72, opened 2020-02-07
[Bug] Visual Studio detects stack corruption in apriltag.c homography_compute2


ScottJohnson2718:
Visual Studio in Debug is telling me that the stack is getting corrupted in apriltag.c in homography_compute2().  See lines 435, 436.

`        double max_val = 0;`
`        int max_val_idx = -1;`

I confirm in the debugger that the invalid index (-1)  for max_val_idx survives to be used as an index later in the function.  So I simply changed the initial value of max_val to be -1.0 and the problem went away because the first iteration of the loop then assigns both values and max_val_idx is no longer -1.

hartter:
I encountered the same issue. Will there be an update of the code that fixes the problem

s-trinh:
@hartter 

Could you attach an image saved in a lossless format (`.png`) to reproduce the issue?

hartter:
Unfortunately I was not able to save the image where this happend (because the process crashed).
Never the less I'd recommend to implement a check, if max_val_idx is -1. This worked for me:

		if (max_val < epsilon) {
			fprintf(stderr, "WRN: Matrix is singular.\n");
		}

		if (max_val_idx == -1) {
			fprintf(stderr, "ERR: Invalid max_val_idx.\n");
			return NULL;
		}

hartter:
> @hartter
> 
> Could you attach an image saved in a lossless format (`.png`) to reproduce the issue?

I originally used TIF but I guess the error will also show up in the PNGs

![Image0](https://user-images.githubusercontent.com/83595887/118779876-c8936580-b88b-11eb-947b-48d38a0d3840.png)
![Image1](https://user-images.githubusercontent.com/83595887/118780016-e95bbb00-b88b-11eb-863c-0d53d35e7230.png)

jrepp:
I was able to reproduce this locally, use image magic to convert to pnm `convert bad-homography.png bad-homography.pnm` then make sure to use the same parameters as python detect `apriltag_demo bad-homography.pnm -x 1` 

The problem as described above is the max index remaining -1 resulting in writes outside of the valid boundary of A later in the function. 

Using a value of -1 is obviously wrong especially when it's multiplied by 9 - this is why we're hitting the guard bytes in stack-protected binaries.  However, I'm not sure what brings the algorithm to a better solution yet. The other trap door to watch out for is the DIV later by A. That is what this check is warning us about. I'm doing some reading and experimenting and will propose a PR that at least prevents both crash scenarios.

[...]

PR #213, opened 2022-02-24
Adding check for length 0 in fit_line to avoid NaN

jrepp:
When dividing by 0 a NaN is introduced to the lineparm, this is eventually introduced to the quads and causes failures later in detection such as quad building.

I tested this with no failing inputs such as https://github.com/AprilRobotics/apriltag/issues/72 as well as verifying I didn't introduce regressions in known good detections.

PR #317, opened 2024-02-04
fix 0 length and determinant

christian-rauch:
This correctly checks for 0 length and determinant and avoids division by zero issues. Supersedes https://github.com/AprilRobotics/apriltag/pull/213 (@jrepp) and fixes https://github.com/AprilRobotics/apriltag/issues/72.


jrepp:
👍

PR #319, opened 2024-03-05
Improve CMakeLists config path by using CMAKE_CURRENT_BINARY_DIR

EwingKang:
Change CMAKE_BINARY_DIR to CMAKE_CURRENT_BINARY_DIR during configuration. This is useful when the project is a subdirectory/ submodule another project (e.g. [apriltags_ros](https://github.com/berndpfrommer/apriltag/tree/ros2)).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,7 +144,7 @@ export(TARGETS apriltag
 
 # install pkgconfig file
 configure_file(${PROJECT_NAME}.pc.in ${PROJECT_NAME}.pc @ONLY)
-install(FILES "${CMAKE_BINARY_DIR}/${PROJECT_NAME}.pc"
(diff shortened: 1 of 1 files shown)


christian-rauch:
Specifically for ROS packages, you should rely on the binary packages or build the library as part of your workspace. But this seems useful nevertheless for projects that do not rely on colcon etc.

christian-rauch:
In this particular case, `apriltag_ros` manually checks out the library in its own source tree. But the correct way to depend on a ROS / colcon package is to specify the dependency in the `package.xml` and use the ROS ecosystem to resolve dependencies in the same way as all other dependencies (`rclcpp`, `cv_bridge`, `image_transport`, ...) are resolved.

PR #332, opened 2024-05-21
apriltag_quad_thresh: Prevent conversion from NaN to int

peci1:
If the user supplied quad_decimate value of 0.0, it works in most places, but in apriltag_quad_thresh, it led to undefined behavior after dividing by this value.

The symptoms are the triciker that on x86/GCC, the undefined behavior leads to -INT_MAX, which gets corrected to 3 pixels by the following lines of code. However, on arm64/GCC, it results in +INT_MAX, which does not get corrected and leads to a running detector which will never detect anything.

### DIFF
--- a/apriltag_quad_thresh.c
+++ b/apriltag_quad_thresh.c
@@ -1810,7 +1810,8 @@ zarray_t* fit_quads(apriltag_detector_t *td, int w, int h, zarray_t* clusters, i
         normal_border |= !family->reversed_border;
         reversed_border |= family->reversed_border;
     }
-    min_tag_width /= td->quad_decimate;
+    if (td->quad_decimate > 1)
+        min_tag_width /= td->quad_decimate;
     if (min_tag_width < 3) {
         min_tag_width = 3;
     }


peci1:
I (now) know that setting quad_decimate to 0.0 is wrong (or at least doesn't make much sense), but the library doesn't check it and the symptoms were so weird I had to provide this fix.

peci1:
Okay, I've updated this PR to use the consistent approach. I don't think there are any more inconsistencies left in the codebase, but you should rather verify that.

So the current behavior is that any value that is 1.0 or lower practically disables the scaling.

christian-rauch (changes requested):
Can you squash those two commits together? It now only makes sense to keep the check for >1 as commit with some matching commit message.

peci1:
Squashed.

christian-rauch:
> Squashed.

The commit is still using the old commit message that does not make sense any more with the squashed diff. Can you update the commit message to something that fits the diff better?

peci1:
It seemed to me the commit message still applies. I kept everything that was there and added one more paragraph explaining that also other values lower than 1.0 are affected by this commit.

If you want something else, please, propose the commit message.

christian-rauch:
> It seemed to me the commit message still applies. I kept everything that was there and added one more paragraph explaining that also other values lower than 1.0 are affected by this commit.

The commit message is still talking about division-by-zero and the effect on different target architectures for GCC. But I think this is not what this commit is about. Yes, it prevents division-by-zero but mainly it aligns the behaviour for the `quad_decimate` value range across the code.

> If you want something else, please, propose the commit message.

I would probably write something like "prevent using decimate for scale smaller 1" or "prevent upscaling of min tag width" as the summary and then just refer to how this aligns with the behaviour in the remaining usages of `quad_decimate`.

peci1:
Ok, what about now?

christian-rauch (approved):

christian-rauch:
> Ok, what about now?

Thanks!

PR #337, opened 2024-06-22
Use _WIN32 macro everywhere

s-trinh:
Looks like a typo to me.

---

There is no `WIN32` macros (at least with usual compilers):
- http://jdebp.info/FGA/predefined-macros-platform.html
- https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170
- https://sourceforge.net/p/predef/wiki/OperatingSystems/
- http://docwiki.embarcadero.com/RADStudio/Athens/en/Predefined_Macros

### DIFF
--- a/common/workerpool.c
+++ b/common/workerpool.c
@@ -198,7 +198,7 @@ void workerpool_run(workerpool_t *wp)
 
 int workerpool_get_nprocs()
 {
-#ifdef WIN32
+#ifdef _WIN32
     SYSTEM_INFO sysinfo;
     GetSystemInfo(&sysinfo);
     return sysinfo.dwNumberOfProcessors;


christian-rauch:
Are you sure that the `WIN32` macro does not exist? It may look like a typo because the other macros are named `_WIN32`, but `WIN32` is indeed a defined macro.

This is definitely not a typo, since the macro exists. But I also find it confusing using `WIN32` while all other Windows-specific sections of the code use `_WIN32`.

Can you rephrase your commit message to make clear that this is not a typo but rather to use a common macro or something similar?

s-trinh:
I have updated the commit accordingly. Since it could be specific to MSVC, it is better to keep the original code IMO.

Luckily we have online tool...
https://godbolt.org/z/8EKTj3sGc

common/workerpool.c
```diff
@@ -198,7 +198,7 @@ void workerpool_run(workerpool_t *wp)
 
 int workerpool_get_nprocs()
 {
-#ifdef WIN32
+#ifdef WIN32 // "WIN32" is defined through <windows.h> header
```
christian-rauch, line 201: I thought you wanted to use the `_WIN32` macro here to be consistent with the rest of the code base. Do you want to replace the `WIN32` macro of keep it? I am now in favour of using `_WIN32` consistently.

christian-rauch (changes requested):

christian-rauch:
> I have updated the commit accordingly. Since it could be specific to MSVC, it is better to keep the original code IMO.

Both macros are defined. But as you pointed out, this looks like a typo since all other Windows-specific sections use a different macro. I prefer to use the same macro for all OS-specific sections to avoid confusion.

common/workerpool.c
s-trinh: Are we sure `SYSTEM_INFO sysinfo;` and following lines could not be specific to a toolchain or something like that?

And it was the intend of the original author to use this specific macro?

---

Otherwise, sure I can update the MR.

christian-rauch: > And it was the intend of the original author to use this specific macro?

The use of the macros originates from https://github.com/AprilRobotics/apriltag/pull/27. There is no mention of explicitly using the two different macros.

> Are we sure `SYSTEM_INFO sysinfo;` and following lines could not be specific to a toolchain or something like that?

The function [`GetSystemInfo`](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getsysteminfo) also seems to be part of the standard Win32 API.

The CI is testing the GCC, clang and MSVC compilers on Windows. If none of them complains, then I think we are pretty safe in using that macro.

christian-rauch (approved):

Issue #70, opened 2020-01-21
Crititical Run in workpool


pmsoftware78:
Hi All,
 I suggest to wait until all threads completed startup before returning in workerpool_create function, adding

```
        pthread_mutex_lock(&wp->mutex);

        while (wp->end_count < wp->nthreads) {
            pthread_cond_wait(&wp->endcond, &wp->mutex);
        }
        
        pthread_mutex_unlock(&wp->mutex);
```

and in workerpool_run function, wp->end_count editing should be protected (ok... this problem should not happen), inserting it after mutex lock:

```
        pthread_mutex_lock(&wp->mutex);
        wp->end_count = 0; 
        pthread_cond_broadcast(&wp->startcond);
```

Best
Paolo

mkrogius:
Hi Paolo, I'm not sure I see exactly what the problem is with the current code. What race condition are you seeing?

pmsoftware78:
Hi Max,
 urgh... so time ago! I try to remember all the details....
 during workerpool_create all threads are created using pthread_create but not all threads at the end of this function can be started and/or not all threads are blocked on pthread_cond_wait, so this part of the code in worker_thread

```
pthread_mutex_lock(&wp->mutex);
        while (wp->taskspos == zarray_size(wp->tasks)) {
            wp->end_count++;  //< this 
```

and 

```
void workerpool_run(workerpool_t *wp)
{
    if (wp->nthreads > 1) {
        wp->end_count = 0; //< this
```

could be execute in wrong order and  wp->end_count will have wrong value.

I try to understand what you want to do and my idea is to wait that all threads be in a wait state

```
  pthread_mutex_lock(&wp->mutex);

        while (wp->end_count < wp->nthreads) {
            pthread_cond_wait(&wp->endcond, &wp->mutex);
        }
        
        pthread_mutex_unlock(&wp->mutex);
```
at the end of create and for extra security lock endcound access
```
// runs all added tasks, waits for them to complete.
void workerpool_run(workerpool_t *wp)
{
    if (wp->nthreads > 1) {

        pthread_mutex_lock(&wp->mutex);
        wp->end_count = 0; // FIX
        pthread_cond_broadcast(&wp->startcond);

        while (wp->end_count < wp->nthreads) {
//            printf("caught %d\n", wp->end_count);
            pthread_cond_wait(&wp->endcond, &wp->mutex);
        }

        pthread_mutex_unlock(&wp->mutex);

        wp->taskspos = 0;

[...]

PR #339, opened 2024-06-24
Fix spurious wake-ups and worker threads initialization in workerpool

berteauxjb:
I encountered random segfaults while using this library and tried to figure out what was happening. I isolated the segfault to the function `worker_thread()` of `workerpool.c`. Long story short, there is no predicate that prevents `worker_thread()` from running if it is woken up spuriously and a task has already been added. Since there is no mutex used in `workerpool_add_task()`, it is possible for the zarray `wp->tasks` to be reallocated, rendering the volatile pointer `task` in `worker_thread()` invalid and causing a segfault.

This PR addresses this issue by adding a predicate and shielding calls that access `wp` behind a mutex. I also saw issue #70 and agree that the initialization of the worker threads might lead to race conditions, so I addressed that too.

### DIFF
--- a/common/workerpool.c
+++ b/common/workerpool.c
@@ -117,6 +117,13 @@ workerpool_t *workerpool_create(int nthreads)
                 return NULL;
             }
         }
+
+        // Wait for the worker threads to be ready
+        pthread_mutex_lock(&wp->mutex);
+        while (wp->end_count < wp->nthreads) {
+            pthread_cond_wait(&wp->endcond, &wp->mutex);
+        }
+        pthread_mutex_unlock(&wp->mutex);
     }
 
     return wp;
--- a/common/workerpool.c
+++ b/common/workerpool.c
@@ -30,6 +30,7 @@ either expressed or implied, of the Regents of The University of Michigan.
 #define __USE_GNU
 #include "common/pthreads_cross.h"
 #include <assert.h>
+#include <stdbool.h>
 #include <stdio.h>
 #include <stdlib.h>
 #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


christian-rauch:
Thanks for the PR. I approved the CI for this PR. The tests crash on Windows (`Exception: SegFault`). Any idea what is going on there?

berteauxjb:
> Thanks for the PR. I approved the CI for this PR. The tests crash on Windows (`Exception: SegFault`). Any idea what is going on there?

I was able to reproduce on a Windows machine :smile: It turns out that the `wp->mutex` is only initialized if `wp->nthreads > 1`. In the test there is only 1 thread, and `workerpool_add_task` was accessing a non-initialized mutex. I shielded the usage of the mutex with `if (wp->nthreads > 1)` like it's done in other places in the file.

christian-rauch (changes requested):
Final request: Can you now squash your fixup commit https://github.com/AprilRobotics/apriltag/pull/339/commits/737ff6b80c32dadb8a470cf38e113f9dffab204b into the original commit https://github.com/AprilRobotics/apriltag/pull/339/commits/be17a2d667f668b05944cfa0e8577e3f262babdc and force-push again with the clean git history?

berteauxjb:
> Final request: Can you now squash your fixup commit [737ff6b](https://github.com/AprilRobotics/apriltag/commit/737ff6b80c32dadb8a470cf38e113f9dffab204b) into the original commit [be17a2d](https://github.com/AprilRobotics/apriltag/commit/be17a2d667f668b05944cfa0e8577e3f262babdc) and force-push again with the clean git history?

Done!

christian-rauch (approved):

christian-rauch:
Thanks a lot!

PR #346, opened 2024-09-02
fix refine_edges() and the coordinates recovery when quad_decimate > …

### LINKED ISSUE #334: corner location has always 0.125 pixel error
**Describe the bug**
i generate a apriltag 36h11 pic to test the corner locations by apriltag lib and theory,but cannot match

**To Reproduce**
here is my pic
[arucoBoard_DICT-DICT_APRILTAG_36h11.zip](https://github.com/user-attachments/files/15612579/arucoBoard_DICT-DICT_APRILTAG_36h11.zip)

**Expected behavior**
**if the corner orign(up-left) of lib is (0,0),the corner should be (9.5,9.5)
if the corner orign(up-left) of lib is (0.5,0.5),the corner should be (10,10)**
**but the output is (9.875,9.875)**

** Input Image**
image
image

** Code version **
3.4.0 release

### LINKED ISSUE #345: About the convention of the image coordinate: the coordinate recovery when quad_decimate > 1 might be wrong?
I suppose we're using the coordinate convention mentioned in this [link](https://answers.opencv.org/question/87923/sub-pixel-coordinate-origin/) in apriltag, where the origin of the image coordinate is defined at the "center," rather than the left-top corner, of the initial pixel.

This convention aligns with most parts of apriltag's implementation: the coordinates of the black/white pixels (centers) are all represented using integers.

However, the [code](https://github.com/AprilRobotics/apriltag/blob/3806edf38ac4400153677e510c9f9dcb81f472c8/apriltag.c#L1097) for recovering coordinates from decimated images contradicts the convention:

```
                    q->p[j][0] = (q->p[j][0] - 0.5)*td->quad_decimate + 0.5;
                    q->p[j][1] = (q->p[j][1] - 0.5)*td->quad_decimate + 0.5;
```

According to the convention mentioned above, the left-top corner of the first pixel should always be (-0.5, -0.5). However, the code above does not map (-0.5, -0.5) to (-0.5, -0.5).
Could this be a typo? If we reverse the signs of all 0.5s in the code, we achieve the expected mapping. Therefore, I suggest the corrected code should be:

```
                    q->p[j][0] = (q->p[j][0] + 0.5)*td->quad_decimate - 0.5;
                    q->p[j][1] = (q->p[j][1] + 0.5)*td->quad_decimate - 0.5;
```

It is straightforward to verify that (-0.5, -0.5) is a (the) fixed point under the new mapping.

NewThinker-Jeffrey:
fix refine_edges() and the coordinates recovery when quad_decimate > 1 (fixes #334, fixes #345)

1. Fix the inconsistent image coordinate conventions used in the code: the convention that (0,0) be the left top corner of the first pixel is adopted by this PR.
2. Fix refine_edges(): Use continuous coordinates and bilinear interpolation (instead of discrete coordinates) when retrieving grayscale values from the input image. Using discrete coordinates might incur extra errors thus unstable.

### DIFF
--- a/apriltag.c
+++ b/apriltag.c
@@ -789,10 +789,18 @@ static void refine_edges(apriltag_detector_t *td, image_u8_t *im_orig, struct qu
             // search on another pixel in the first place. Likewise,
             // for very small tags, we don't want the range to be too
             // big.
-            double range = td->quad_decimate + 1;
+
+            int range = td->quad_decimate + 1;
+
+            // To reduce the overhead of bilinear interpolation, we can
+            // reduce the number of steps per unit.
+            int steps_per_unit = 4;
+            double step_length = 1.0 / steps_per_unit;
+            int max_steps = 2 * steps_per_unit * range + 1;
 
             // XXX tunable step size.
-            for (double n = -range; n <= range; n +=  0.25) {
+            for (int step = 0; step < max_steps; ++step) {
+                double n = -range + step_length * step;
                 // Because of the guaranteed winding order of the
                 // points in the quad, we will start inside the white
                 // portion of the quad and work our way outward.
@@ -802,19 +810,36 @@ static void refine_edges(apriltag_detector_t *td, image_u8_t *im_orig, struct qu
                 // gradient more precisely, but are more sensitive to
                 // noise.
                 double grange = 1;
-                int x1 = x0 + (n + grange)*nx;
-                int y1 = y0 + (n + grange)*ny;
-                if (x1 < 0 || x1 >= im_orig->width || y1 < 0 || y1 >= im_orig->height)
+
+                double x1 = x0 + (n + grange)*nx;
+                double y1 = y0 + (n + grange)*ny;
+                double x1i_d, y1i_d, a1, b1;
+                a1 = modf(x1, &x1i_d);
+                b1 = modf(y1, &y1i_d);
+                int x1i = x1i_d, y1i = y1i_d;
+
+                if (x1i < 0 || x1i + 1 >= im_orig->width || y1i < 0 || y1i + 1 >= im_orig->height)
                     continue;
 
-                int x2 = x0 + (n - grange)*nx;
-                int y2 = y0 + (n - grange)*ny;
-                if (x2 < 0 || x2 >= im_orig->width || y2 < 0 || y2 >= im_orig->height)
+                double x2 = x0 + (n - grange)*nx;
+                double y2 = y0 + (n - grange)*ny;
+                double x2i_d, y2i_d, a2, b2;
+                a2 = modf(x2, &x2i_d);
+                b2 = modf(y2, &y2i_d);
+                int x2i = x2i_d, y2i = y2i_d;
+
+                if (x2i < 0 || x2i + 1 >= im_orig->width || y2i < 0 || y2i + 1 >= im_orig->height)
                     continue;
 
-                int g1 = im_orig->buf[y1*im_orig->stride + x1];
-                int g2 = im_orig->buf[y2*im_orig->stride + x2];
-
+                // interpolate
+                double g1 = (1 - a1) * (1 - b1) * im_orig->buf[y1i*im_orig->stride + x1i] +
+                                  a1 * (1 - b1) * im_orig->buf[y1i*im_orig->stride + x1i + 1] +
+                            (1 - a1) *    b1    * im_orig->buf[(y1i + 1)*im_orig->stride + x1i] +
+                                  a1 *    b1    * im_orig->buf[(y1i + 1)*im_orig->stride + x1i + 1];
+                double g2 = (1 - a2) * (1 - b2) * im_orig->buf[y2i*im_orig->stride + x2i] +
+                                  a2 * (1 - b2) * im_orig->buf[y2i*im_orig->stride + x2i + 1] +
+                            (1 - a2) *    b2    * im_orig->buf[(y2i + 1)*im_orig->stride + x2i] +
+                                  a2 *    b2    * im_orig->buf[(y2i + 1)*im_orig->stride + x2i + 1];
                 if (g1 < g2) // reject points whose gradient is "backwards". They can only hurt us.
                     continue;
 
--- a/apriltag.c
+++ b/apriltag.c
@@ -797,6 +797,7 @@ static void refine_edges(apriltag_detector_t *td, image_u8_t *im_orig, struct qu
             int steps_per_unit = 4;
             double step_length = 1.0 / steps_per_unit;
             int max_steps = 2 * steps_per_unit * range + 1;
+            double delta = 0.5;
 
             // XXX tunable step size.
             for (int step = 0; step < max_steps; ++step) {
@@ -811,8 +812,8 @@ static void refine_edges(apriltag_detector_t *td, image_u8_t *im_orig, struct qu
                 // noise.
                 double grange = 1;
 
-                double x1 = x0 + (n + grange)*nx;
-                double y1 = y0 + (n + grange)*ny;
+                double x1 = x0 + (n + grange)*nx - delta;
+                double y1 = y0 + (n + grange)*ny - delta;
                 double x1i_d, y1i_d, a1, b1;
                 a1 = modf(x1, &x1i_d);
                 b1 = modf(y1, &y1i_d);
@@ -821,8 +822,8 @@ static void refine_edges(apriltag_detector_t *td, image_u8_t *im_orig, struct qu
                 if (x1i < 0 || x1i + 1 >= im_orig->width || y1i < 0 || y1i + 1 >= im_orig->height)
                     continue;
 
-                double x2 = x0 + (n - grange)*nx;
-                double y2 = y0 + (n - grange)*ny;
+                double x2 = x0 + (n - grange)*nx - delta;
+                double y2 = y0 + (n - grange)*ny - delta;
                 double x2i_d, y2i_d, a2, b2;
                 a2 = modf(x2, &x2i_d);
                 b2 = modf(y2, &y2i_d);
@@ -1115,13 +1116,8 @@ zarray_t *apriltag_detector_detect(apriltag_detector_t *td, image_u8_t *im_orig)
             zarray_get_volatile(quads, i, &q);
 
             for (int j = 0; j < 4; j++) {
-                if (td->quad_decimate == 1.5) {
-                    q->p[j][0] *= td->quad_decimate;
-                    q->p[j][1] *= td->quad_decimate;
-                } else {
-                    q->p[j][0] = (q->p[j][0] - 0.5)*td->quad_decimate + 0.5;
-                    q->p[j][1] = (q->p[j][1] - 0.5)*td->quad_decimate + 0.5;
-                }
+                q->p[j][0] *= td->quad_decimate;
+                q->p[j][1] *= td->quad_decimate;
             }
         }
     }
--- a/apriltag_detect.docstring
+++ b/apriltag_detect.docstring
@@ -42,7 +42,9 @@ a tuple containing the detections. Each detection is a dict with keys:
 
 - id: integer identifying each detected tag
 
-- center: pixel coordinates of the center of each detection
+- center: pixel coordinates of the center of each detection.  NOTE: Please be
+  cautious regarding the image coordinate convention. Here, we define (0,0) as
+  the left-top corner (not the center point) of the left-top-most pixel.
 
 - lb-rb-rt-lt: pixel coordinates of the 4 corners of each detection. The order
   is left-bottom, right-bottom, right-top, left-top


apriltag.c
```diff
@@ -789,10 +789,22 @@ static void refine_edges(apriltag_detector_t *td, image_u8_t *im_orig, struct qu
...
             // for very small tags, we don't want the range to be too
             // big.
-            double range = td->quad_decimate + 1;
+
+            int range = td->quad_decimate + 1;
+            
+            // to reduce the overhead of bilinear interpolation, we can
+            // reduce the number of steps per unit.
+            int steps_per_unit = 4;
+            // int steps_per_unit = 2;
```
christian-rauch, line 798: Do not commit commented-out code.

apriltag.c
```diff
@@ -789,10 +789,22 @@ static void refine_edges(apriltag_detector_t *td, image_u8_t *im_orig, struct qu
...
+
+            double step_length = 1.0 / steps_per_unit;
+            int max_steps = 2 * steps_per_unit * range + 1;
+            double delta = 0.5;
 
             // XXX tunable step size.
-            for (double n = -range; n <= range; n +=  0.25) {
+            // for (double n = -range; n <= range; n +=  0.25) {
```
christian-rauch, line 805: Do not commit commented-out code.

apriltag.c
```diff
@@ -802,24 +814,39 @@ static void refine_edges(apriltag_detector_t *td, image_u8_t *im_orig, struct qu
...
+                                  a1 *    b1    * im_orig->buf[(y1i + 1)*im_orig->stride + x1i + 1];
+                double g2 = (1 - a2) * (1 - b2) * im_orig->buf[y2i*im_orig->stride + x2i] +
+                                  a2 * (1 - b2) * im_orig->buf[y2i*im_orig->stride + x2i + 1] +
+                            (1 - a2) *    b2    * im_orig->buf[(y2i + 1)*im_orig->stride + x2i] +
+                                  a2 *    b2    * im_orig->buf[(y2i + 1)*im_orig->stride + x2i + 1];
                 if (g1 < g2) // reject points whose gradient is "backwards". They can only hurt us.
                     continue;
 
                 double weight = (g2 - g1)*(g2 - g1); // XXX tunable. What shape for weight=f(g2-g1)?
-
```
christian-rauch, line 822: Keep the empty line to separate the comment from the previous code block.

apriltag.c
```diff
@@ -1094,8 +1121,14 @@ zarray_t *apriltag_detector_detect(apriltag_detector_t *td, image_u8_t *im_orig)
                     q->p[j][0] *= td->quad_decimate;
                     q->p[j][1] *= td->quad_decimate;
                 } else {
-                    q->p[j][0] = (q->p[j][0] - 0.5)*td->quad_decimate + 0.5;
-                    q->p[j][1] = (q->p[j][1] - 0.5)*td->quad_decimate + 0.5;
+                    // q->p[j][0] = (q->p[j][0] - 0.5)*td->quad_decimate + 0.5;
+                    // q->p[j][1] = (q->p[j][1] - 0.5)*td->quad_decimate + 0.5;
```
christian-rauch, line 1125: Do not commit commented-out code.

apriltag.c
```diff
@@ -1094,8 +1121,14 @@ zarray_t *apriltag_detector_detect(apriltag_detector_t *td, image_u8_t *im_orig)
...
+
+                    q->p[j][0] = q->p[j][0] * td->quad_decimate;
+                    q->p[j][1] = q->p[j][1] * td->quad_decimate;
+
+                    // q->p[j][0] = (q->p[j][0] + 0.5)*td->quad_decimate - 0.5;
+                    // q->p[j][1] = (q->p[j][1] + 0.5)*td->quad_decimate - 0.5;
```
christian-rauch, line 1131: Do not commit commented-out code.

christian-rauch (changes requested):
I only left style comments so far as I need a while to better understand the changes. Can you add more details about the issue and how it is fixed in the commit message?

christian-rauch:
Can you add a qualitative and quantitative comparison of the estimation results before and after your PR?

NewThinker-Jeffrey:
> Can you add a qualitative and quantitative comparison of the estimation results before and after your PR?

Of course, let me make it clear.

I tested with [this image](https://github.com/user-attachments/files/15612579/arucoBoard_DICT-DICT_APRILTAG_36h11.zip) from #334 .
The ground truth coordinate for the left top corner of the first tag is (10,10), as shown in #334 :
image

**With the current master branch**, the estimated coordinates of that corner under different settings are:
(1). `refine_edges=false, quad_decimate=1`:  `(9.997555, 10.000000)` 
(2). `refine_edges=false, quad_decimate=2`:   `(9.500000, 9.500000)`  
(3). `refine_edges=true, quad_decimate=1`:   `(10.130014, 9.875005)`  
(4). `refine_edges=true, quad_decimate=2`:   `(9.875000, 9.875005)`   (This is the default setting of Apriltag, and the result coincides with that in #334 )

As we can see,
- The master branch produces correct estimation only under setting (1);
- While under setting (2), an error of `(0.5, 0.5)` is present due to the inconsistent coordinate conventions used in the code, which can be fixed by a small modification [here](https://github.com/NewThinker-Jeffrey/apriltag/blob/7f64834abfc26903b63dee962edca995ec232368/apriltag.c#L1123);
- If `refine_edges` is enabled (setting (3) and (4)), extra error can be introduced which is mainly caused by the discrete coordinates used when retrieving grayscale values from the input image. This can be addressed by applying continuous coordinates and bilinear interpolation.

**With the modified code** (this PR), the estimated coordinates under different settings are all close to the groundtruth (10,10):
(1). `refine_edges=false, quad_decimate=1`:  `(9.997555, 10.000000)`
(2). `refine_edges=false, quad_decimate=2`:   `(10.000000, 10.000000)`  
(3). `refine_edges=true, quad_decimate=1`:   `(9.999912, 10.000005)`  
(4). `refine_edges=true, quad_decimate=2`:   `(10.000000, 10.000005)`

apriltag.c
NewThinker-Jeffrey: A new commit has addressed this.

apriltag.c
```diff
@@ -1094,8 +1119,9 @@ zarray_t *apriltag_detector_detect(apriltag_detector_t *td, image_u8_t *im_orig)
                     q->p[j][0] *= td->quad_decimate;
                     q->p[j][1] *= td->quad_decimate;
                 } else {
-                    q->p[j][0] = (q->p[j][0] - 0.5)*td->quad_decimate + 0.5;
-                    q->p[j][1] = (q->p[j][1] - 0.5)*td->quad_decimate + 0.5;
+                    // Maybe merged with the above if? They are doing the same thing now.
```
NewThinker-Jeffrey, line 1122: @christian-rauch  Please consider this comment.

christian-rauch: Do you mean merging the `td->quad_decimate == 1.5` if/else blocks into one?

christian-rauch (changes requested):
Can you squash your commits together again?

christian-rauch:
Thanks to this, I found that I broke the tests with https://github.com/AprilRobotics/apriltag/pull/327 when I tried to make them less flaky. The test compares the pixel coordinates and should trigger an error if the absolute error is larger than 0.1 pixels:
https://github.com/AprilRobotics/apriltag/blob/3806edf38ac4400153677e510c9f9dcb81f472c8/test/test_detection.c#L47-L54

Setting:
```C++
td->quad_decimate = 2;
td->refine_edges = false;
```
in the tests should therefore have triggered this when running `ctest --test-dir test/`.

I fixed the test on Ubuntu in https://github.com/AprilRobotics/apriltag/pull/348. But on other OSes, the results are slightly different and therefore have a small discrepancy to the reference values:
```
Got:
  0, (357.3853 346.3749), (373.2110 347.5041), (373.4671 329.2181), (357.2913 328.1352)
Expected:
  0, (357.4018 346.3447), (373.2110 347.5034), (373.4671 329.2181), (357.1758 328.1275)
```
and I am wondering if this may be caused by rounding effects and fixed by this PR.

christian-rauch:
I changed the tests to not decimate quads or refine edges:
```C++
    td->quad_decimate = 1;
    td->refine_edges = false;
```
and now get identical (accurate up to 0.001 pixels) results across OSes, which suggests that decimation and refinement are causing unreliable detections.

While `refine_edges=false, quad_decimate=1` is supposed to give the true stable solution, I get different results for decimation and refinement
```C++
    td->quad_decimate = 2;
    td->refine_edges = true;
```
after applying your first commit.

Can you show with debug results from the test images, that this PR should produce the same quads, just better refined?

apriltag.c
NewThinker-Jeffrey: Yes. I'll merge the two blocks in the next commit if you think it's ok too.

NewThinker-Jeffrey:
> Can you show with debug results from the test images, that this PR should produce the same quads, just better refined?

I'm not entirely sure what you mean by "debug results from the test images". I have simply enabled `td->debug` and tested the setting `refine_edges=true, quad_decimate=2` again with [the image](https://github.com/user-attachments/files/15612579/arucoBoard_DICT-DICT_APRILTAG_36h11.zip) mentioned above. The debug output has been packed into a .zip file, please refer to the attachment.
[apriltag_debug.zip](https://github.com/user-attachments/files/16860486/apriltag_debug.zip)

NewThinker-Jeffrey:
> Can you squash your commits together again?

I'll squash the two commits into one after you confirm whether we should merge the if/else blocks mentioned above.

christian-rauch:
> It can be observed that master and this pr produce slightly different coordinates for the tag corners.

Thanks for the comparison. The results are indeed very close.

> However it's hard to tell which version is better than the other according to these results only, because the ground truths for real world images are unknown.

I selected real images since some processing steps, such as the edge refinement, will have more noticeable effects on noisy real images than clean synthetic images.

> Anyway, theoretically speaking, using continuous coordinates (with interpolation) to retrieve pixel values, as done in this pr, can generate more accurate edge points, and the test with the ideal input image (from #334 ) , where ground truth is available, does show that this pr can produce nearly perfect estimation for the corners' coordinates.

I originally added the test images to see if a PR would change the estimated quads. For a performance evaluation of which method is better, we would need to manually label these images and extract ground truth corner coordinates. Alternatively, we could render synthetic images for the ground truth coordinates and add blur, noise and other effects to see how a change affects the performance.

apriltag.c
NewThinker-Jeffrey: The if/else blocks are merged in the latest (squashed) commit.

NewThinker-Jeffrey:
@christian-rauch  Hi, I have squashed the previous commits. Please let me know if any further modifications are needed.

christian-rauch:
`fixes #334, fixes #345` is not a nice commit message title. The commit message title should summarise what the change is and does and the commit message body should have the details. Your commit has quite a few functional changes. Documenting the effect and why this is done is therefore important. Your commit message body contains two bullet points. Is it possible to split this one commit into two separate functional changes, i.e. one that fixes the image coordinate conventions and another one that touches the bilinear interpolation?

I recently fixed the tests to work across OSes with no decimation. Can you rebase your PR to make sure that it does not break the tests without decimation? Then, we probably need a test for the results with decimation to make sure that results with decimation are close to the ones without decimation and that the output does not change in future.

NewThinker-Jeffrey:
Hi christian-rauch,

> Is it possible to split this one commit into two separate functional changes, i.e. one that fixes the image coordinate conventions and another one that touches the bilinear interpolation?

Yes, it's better to split the commit into two.

I think we should update the doc file `apriltag_detect.docstring` or `README.md` to add an explicit explanation to the image coordinate convention somewhere. For example in `apriltag_detect.docstring`:

```
- center: pixel coordinates of the center of each detection. NOTE: Please be
  cautious regarding the image coordinate convention. Here, we define (0,0) as
  the top-left corner (not the center point) of the top-leftmost pixel.

- lb-rb-rt-lt: pixel coordinates of the 4 corners of each detection. The order
  is left-bottom, right-bottom, right-top, left-top.
```

Or maybe you have better suggestions?

apriltag.c
```diff
@@ -802,19 +810,35 @@ static void refine_edges(apriltag_detector_t *td, image_u8_t *im_orig, struct qu
...
                 double grange = 1;
-                int x1 = x0 + (n + grange)*nx;
-                int y1 = y0 + (n + grange)*ny;
-                if (x1 < 0 || x1 >= im_orig->width || y1 < 0 || y1 >= im_orig->height)
+                double x1 = x0 + (n + grange)*nx;
+                double y1 = y0 + (n + grange)*ny;
+                int x1i = floor(x1);
+                int y1i = floor(y1);
+                double a1 = x1 - x1i;
+                double b1 = y1 - y1i;
```
christian-rauch, line 819: Do you just want to get the integral and fractional part? If so, just use [`modf`](https://en.cppreference.com/w/c/numeric/math/modf).

apriltag.c
```diff
@@ -802,19 +810,35 @@ static void refine_edges(apriltag_detector_t *td, image_u8_t *im_orig, struct qu
...
-                int x2 = x0 + (n - grange)*nx;
-                int y2 = y0 + (n - grange)*ny;
-                if (x2 < 0 || x2 >= im_orig->width || y2 < 0 || y2 >= im_orig->height)
-                    continue;
+                double x2 = x0 + (n - grange)*nx;
+                double y2 = y0 + (n - grange)*ny;
+                int x2i = floor(x2);
+                int y2i = floor(y2);
+                double a2 = x2 - x2i;
+                double b2 = y2 - y2i;
```
christian-rauch, line 829: Same question: What's the purpose of rounding down here?

christian-rauch (changes requested):
Thank you for splitting this. I think the purpose of the diffs are much easier to understand.

Regarding the commits message: I think the title is far too long and the message body also should have line breaks. If you look at an individual commit, e.g. https://github.com/AprilRobotics/apriltag/pull/346/commits/5666602a8a115bfae2f0991acd99fff67ba490c7, then ideally the title fits into the first line without line breaks.

Have a look at https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/ for suggestions on how to format a good commit message regarding style and content.

NewThinker-Jeffrey:
> Thank you for splitting this. I think the purpose of the diffs are much easier to understand.
> 
> Regarding the commits message: I think the title is far too long and the message body also should have line breaks. If you look at an individual commit, e.g. [5666602](https://github.com/AprilRobotics/apriltag/commit/5666602a8a115bfae2f0991acd99fff67ba490c7), then ideally the title fits into the first line without line breaks.

Modified as you suggested, please see the new commits.

christian-rauch (approved):
Thanks. I am fine with these changes now. But I still would like @mkrogius to have a look at this and give the green light.

christian-rauch:
As said, I am fine with these changes now. I would still like @mkrogius to have a look at this.

@NewThinker-Jeffrey What is your deadline on this? You can create a new branch from this and start working there. You do not have to (and you shouldn't) push new changes to this PR.

NewThinker-Jeffrey:
> As said, I am fine with these changes now. I would still like @mkrogius to have a look at this.
> 
> @NewThinker-Jeffrey What is your deadline on this? You can create a new branch from this and start working there. You do not have to (and you shouldn't) push new changes to this PR.

I understand that and won't push my new changes to this branch before the PR finished.

christian-rauch:
I am going to merge this now to not delay this further and potentially causing merge conflicts later.

Issue #355, opened 2024-09-26
Porting to single core embedded


Chris-F5:
For my application, I'd like to detect april tags on a single core of an RP2040. This means there is no `<pthread.h>` for me to include and use. I'm happy to try and implement this port myself, but I create this issue for 3 reasons:

1. Has anybody attempted this before? Is there some existing port I haven't been able to find?
2. Because I'm open to input on how the best way to implement this would be.
3. Is this something that would be usefull to have in the main repo. I.e. should I make a pull request once I'm done?

At first glance, it seems that a good way to do this would be to add a new macro option in `common/pthreads_cross.{c,h}`, similar to `_WIN32`. Something like:
```
#ifdef _NOTHREADS
typedef void pthread_mutex_p;
...
```
And then add some checks that only one thread is being created elsewhere. By doing this I would be eliminating all dependence on `<pthread.h>` when `_NOTHREADS` is set.

christian-rauch:
I would be up for making threading optional via a CMake option. If threading is deactivated, then everything should behave via `#ifdef`s as if a single thread was created.

Chris-F5:
[Here](https://github.com/Chris-F5/apriltag/commit/0ce68e62af5e14d86b3c8e63133c6f395dd2b05c) is my NOTHREADS macro implementation. If `NOTHREADS` is defined then `pthreads_cross.{c,h}` will declare and define `pthread_*` functions which return immediately (this is just so that all symbols will get resolved). `workerpool.c` will then enforce a single thread. I'm not so great with CMake so I haven't added an option in that yet.

Some other things I noticed when trying to make my rp2040 micropython port:

- `pthreads_cross.{c.h}` were using CRLF line ends. (pull request created)
- `int32_t` types are being foratted with `%d` which assumed that `int` is 32 bits. `int32_t` types should be formatted with `PRId32`. See the [docs](https://en.cppreference.com/w/c/types/integer). I have fixed some of these in [this commit](https://github.com/Chris-F5/apriltag/commit/e6ce29b3dfec6ba5033101ee2f90e3fbf201cf72) but its possible I missed some, since I'm not using the full source tree. I can create a pull request if you want.

If you need to detect apriltags on an embedded device running micropython, you can check out [the branch on my fork](https://github.com/Chris-F5/apriltag/tree/micropython).

christian-rauch:
[...]

PR #356, opened 2024-09-30
Remove CRLF line ends

Chris-F5:
I'm doing some work on `common/pthreads_cross.{c,h}` related to #355 . I'm bothered by the CRLF line ends which are used only in these two files. This commit changes nothing except removing the carriage return characters.


christian-rauch:
Are you only bothered by the CRLF or generally by having different file formats and encodings for the source files?

If so, you may want to check:
```sh
find . -type f -name "*.c" -o -name "*.h" -o -name "*.cc" | file -f -
```
which shows `Unicode text, UTF-8 text` instead of `ASCII text` for some of the source files (`./common/matd.c`, `./apriltag_pose.h`) :-)

If we are already touching every line in some files without actual code change, can you also convert the `UTF-8` encoded files to `ASCII` via `iconv -f UTF-8 -t ASCII $input -o $output` and check that there are actually no characters omitted?

Chris-F5:
Oh wow I didn't notice those non-ASCII characters. I wonder what kind of code editor put those in!
There is a cool tool called [unidecode](https://pypi.org/project/Unidecode/) which is useful for automatically replacing these kind of unicode characters with their ASCII equivalent.

PR #358, opened 2024-10-01
Fix README C example

Chris-F5:
The C example tried to load a `.png` file with a method for loading `.pnm` files. This will fail and return `NULL` which later causes a segfault.


christian-rauch:
Good catch. Thanks.

PR #384, opened 2025-05-26
Release Python Global Interpreter Lock During Detection

Jgunde:
**Summary:**
This PR releases Python’s Global Interpreter Lock (GIL) during the CPU-intensive detection stage. This allows other threads in the Python process to run concurrently.

**Details:**
- The GIL prevents multiple Python threads from executing in parallel. If the GIL is not released, detection blocks the entire Python process, starving other threads of CPU.
- On low-resolution images or with high decimate values, detection is fast (milliseconds), so blocking is minimal. With high-res images or low decimate values, detection can take up to a second, making GIL release critical for multi-threaded applications.
- Manual GIL release can be unsafe if C code interacts with Python objects or is not thread-safe. Here, this is not an issue because `apriltag_detector_detect()` is implemented purely in C and does not interact with Python objects.

**References:**
[Python docs: Releasing the GIL from extension code](https://docs.python.org/3/c-api/init.html#releasing-the-gil-from-extension-code)

**Testing:**
Changes are tested and work as intended.

### DIFF
--- a/apriltag_pywrap.c
+++ b/apriltag_pywrap.c
@@ -243,7 +243,11 @@ static PyObject* apriltag_detect(apriltag_py_t* self,
                      .stride = strides[0],
                      .buf    = PyArray_DATA(image)};
 
-    zarray_t* detections = apriltag_detector_detect(self->td, &im);
+    zarray_t *detections;  // Declare detections variable outside of the GIL block
+    Py_BEGIN_ALLOW_THREADS  // Acquire the GIL before running detection
+        detections = apriltag_detector_detect(self->td, &im);
+    Py_END_ALLOW_THREADS  // Release the GIL after detection completes
+
     int N = zarray_size(detections);
 
     if (N == 0 && errno == EAGAIN){


Jgunde:
> I am not familiar with handing the GIL manual. Can this cause issues when you run `apriltag_detect` in Python in multiple threads using the same detector?

@christian-rauch 
I didn't consider this use case. Since detection can run on multiple cores, I assumed it would only be used in a single thread as multithreading wouldn't provide a performance benefit.

I just finished some multithreaded testing and your hypothesis is correct. The program crashed when using the same detector object across multiple threads.

I do think releasing the GIL provides a large enough benefit to warrant its inclusion. Maybe we add a note that the detect method is not thread safe?

For those that do want to use a single detector object across multiple threads, they can use the Python threading module's [lock objects](https://docs.python.org/3/library/threading.html#lock-objects). It's common practice to using locks in this way to synchronize access to shared objects.
Personally, it seems simpler to create a distinct detector object for each thread.

It also seems possible to implement the lock in C using the [CPython lock interface](https://github.com/python/cpython/blob/main/Include/pythread.h).

What do you think? Would updating the [Python documentation](https://github.com/AprilRobotics/apriltag?tab=readme-ov-file#python) noting that the detect method is not thread safe be enough?

Jgunde:
> Thanks for testing this. Just to make sure, it does not crash with the commit just before your PR? Do you have a PoC to reproduce this with an example image?

Correct. It does not crash when using the same detector object across multiple threads.

> Is it possible to add this manual locking inside apriltag_detect?

I think so. It'll probably require 10-30 lines of code. I'll look into this.

Jgunde:
Hi Christian,

I've created a PR [here](https://github.com/AprilRobotics/apriltag/pull/386) that implements a lock and acquires it before running detection. This prevents the crash mentioned earlier in this thread.

Best,
Jacob

PR #386, opened 2025-06-01
Acquire Lock Before Releasing Python GIL

Jgunde:
This PR implements a lock in apriltag_pywrap.c to prevent multiple threads which share the same detector object from calling the detect function, apriltag_detector_detect(), at the same time.

Previously, the Python GIL was released during the call to apriltag_detector_detect(). This would cause the program to crash if multiple threads called this function at the same time. By implementing a lock, only a single thread can call it at a time.

The added functionality was tested using the following Python code:

```Python
from PIL import Image
import threading

import numpy as np
from apriltag import apriltag


image = Image.open("apriltag.png").convert("L")  # Load apriltag image and convert to grayscale.
image_array = np.array(image)

detector = apriltag(
    family="tag36h11",
    threads=4,
    maxhamming=1,
    decimate=1.0,
    blur=0.0,
    refine_edges=True,
    debug=False
)


def worker(thread_id):
    for i in range(100):
        detector.detect(image_array)
    print(f"Thread {thread_id} complete.")


# Create and start multiple threads
threads = []
for i in range(10):
    thread = threading.Thread(target=worker, args=(i,))
    thread.start()
    threads.append(thread)

# Wait for all threads to finish
for thread in threads:
    thread.join()
```

### DIFF
--- a/apriltag_pywrap.c
+++ b/apriltag_pywrap.c
@@ -68,6 +68,7 @@ typedef struct {
 
     apriltag_family_t*   tf;
     apriltag_detector_t* td;
+    PyThread_type_lock   det_lock;
     void (*destroy_func)(apriltag_family_t *tf);
 } apriltag_py_t;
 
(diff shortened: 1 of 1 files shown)


christian-rauch (approved):
Thanks a lot for taking care of the issue.

christian-rauch:
Unfortunately, the signing key for the ROS apt repo expired yesterday and hasn't been updated in the `ros-tooling/setup-ros` action yet. The CI will not pass until this has been resolved.

christian-rauch:
The CI seems to work again. Can you rebase and squash your commits together, and then force push again?

PR #389, opened 2025-06-27
add MXE cross-compilation support with pthread and gettimeofda…

hank5000:
- Added `CheckIncludeFile` to detect presence of pthread.h and sys/time.h
- Ensures the project can be cross-compiled using MXE


christian-rauch (commented):
Can you provide more context what MXE is and why these changes are required? Is it possible to add this to the CI pipeline in order to check that this "MXE support" does not break in future?

christian-rauch:
@hank5000 Is this PR still relevant? Can you have a look at this and address the comments, or close the PR if you don't need this any more?

Issue #343, opened 2024-08-13
quick_decode struct allocates  6Gb memory for  standard52h13 family


kde-baskets-user:
**Describe the bug**
quick_decode_init() allocates too much memory for  the standard52h13 family.  Any workarounds?

**To Reproduce**
Create standard52h13 detector to see allocated memory.

**Expected behavior**
Parsing of 15 bits of useful data  from an image should take more reasonable amount of memory. 

**Operating Sytem**
Ubuntu

** Installation Method**
I built AprilTag from source following the instructions in the README

** Code version **
3.4.2 from GitHub

christian-rauch:
I can confirm the high memory consumption by running `./opencv_demo --camera 2 --family tagStandard52h13` and checking the memory consumption with `top -p $(pidof opencv_demo)`, which reports around 6GB RES consumption.

The `tagStandard52h13` has one of the largest code books (48714 codes). Loading an even larger code book from the `tagCircle49h12` (65535 codes) will consume even more memory (more than 7GB in my case).

> Parsing of 15 bits of useful data from an image should take more reasonable amount of memory.

How do you get to 15 bits? As the name `tagStandard52h13` suggests, this tag family encodes 52 bits with a hamming distance of 13.

If you need to reduce the size of the code book, I simply suggest using a different tag family with fewer encoded bits. You can also [generate](https://github.com/AprilRobotics/apriltag-generation) your own tag family for specific requirements.

kde-baskets-user:
Thanks for your answer.  Since tagStandard52h13 has 48714 different entries, one aprilag code contains 15.6 = log2(48724) bits of information, by "useful data" I meant rigorously defined word "information", sorry for any confusion.  Everything else in 52  bits representation has to do with code correction, I assume.  

Now, while memory depending on ncodes grows only polynomially(and not even quadratically, as far as I see), 6Gb for many embedded applications is prohibitively costly especially for a seemingly modest task of error correction, but I am not expert in this field. 

I am currently evaluating several 2D barcode libraries and this is so far a show-stopper for apriltag. I am going to run some more tests and perhaps will be back with more data on this.

christian-rauch:
> Thanks for your answer. Since tagStandard52h13 has 48714 different entries, one aprilag code contains 15.6 = log2(48724) bits of information, by "useful data" I meant rigorously defined word "information", sorry for any confusion. Everything else in 52 bits representation has to do with error correction, I assume.

[...]

PR #397, opened 2025-08-17
Update README.md

nathanieltagg:
Add notes about cost of larger tag families, as per discussion in issue #343.


christian-rauch:
I am not going to merge this, as I think this is too specific and does not help much to select an optimal tag family for one's application.

But I will leave this open in case other maintainers find this specific information useful.

PR #400, opened 2025-09-26
Fix opencv example readme section

reprojection-calibration:
The opencv example readme section directs users to run the following commands:

```
$ cd examples
$ make opencv_demo
```

The first command does not work because the folder is called "example/" not "examples/" and the second command fails with unknown include errors because there is no build configuration or file or Makefile to actually orchestrate building the opencv demo.

Given that the examples are built by default I propose that the readme is changed to simply instruct people to build and then run `./build/opencv_demo`. This works like  charm and is easy to understand.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,8 +136,7 @@ Note that this library has no external dependencies. Most applications
 will require, at minimum, a method for acquiring images.
 
 See example/opencv_demo.cc for an example of using AprilTag in C++ with OpenCV.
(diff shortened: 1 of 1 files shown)


christian-rauch:
Yes, the command `make opencv_demo` only works for the default make file generator and not with `-GNinja`. I would probably skip this specific build instruction altogether. The `opencv_demo` target is built by default if OpenCV is available. We only need to document how to install the OpenCV dependency and how to start the example executable.

PR #401, opened 2025-10-02
fix: windows crosscompiling

VincentMarnier:
Windows crosscompilation is broken again. This fixes the issue. This is due to the fact that in mingw, `Winsock2.h` is `winsock2.h`, and is case sensitive. Might be `winsock2.h` in windows too, but I don't have a computer running windows to check, sadly.

### DIFF
--- a/.github/workflows/cross-compilation.yml
+++ b/.github/workflows/cross-compilation.yml
@@ -1,0 +1,86 @@
+name: Cross compilation
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+  workflow_dispatch:
+
+  env:
+    DEBIAN_FRONTEND: noninteractive
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        shared_libs: ['ON', 'OFF']
+
+    steps:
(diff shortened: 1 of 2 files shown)


christian-rauch:
Could you also use [`__has_include`](https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005finclude.html) on Windows? Can you add a configuration to the CI to verify that this does not break again?

VincentMarnier:
I'll do that, no problem.
I ignored the __has_include thing! Nice to learn ;)

VincentMarnier:
For the CI part, it is way trickier that expected as I build it on linux and run it on windows.
I don't want to change cmake-multi-platform.yml too much.
How would you like me to do it?

christian-rauch:
The `cmake-multi-platform.yml` loops over the Linux/macOS/Windows runners and runs the same procedure "natively" on the runner. I guess it is a bit complicated to squeeze the cross-compilation in there. I would create a new "cross-compilation" workflow to separate this better.

christian-rauch:
> Will test the CI on my side and reopen when ready, don't want to spam you

You can also open PRs in "Draft" state. I think that this does not send out notifications unless you switch it to "Ready".

PR #402, opened 2025-10-02
fix: windows crosscompilation and added a workflow to test it by crosscompiling on ubuntu and running tests on windows

VincentMarnier:
When crosscompiling for windows, `Winsock2.h` is not found, because the `include` statement is case sensitive and the header's names is `winsock2.h` instead of `Winsock2.h`.

We now checks which is present and use it.

Also added a workflow to test the cross compilation by building (using mingw) on ubuntu and running the tests on windows.

### DIFF
--- a/.github/workflows/cross-compilation.yml
+++ b/.github/workflows/cross-compilation.yml
@@ -1,0 +1,87 @@
+name: Cross compilation
+
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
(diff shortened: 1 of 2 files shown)


VincentMarnier:
So here we are again, with the `__has_include` and the CI.
Let me know if someone feels wrong to you!

christian-rauch (commented):
Looks good. Can you maybe just add some info in the commit message and the PR what is happening here and why this is required?

christian-rauch (approved):

PR #403, opened 2025-10-07
Avoid malloc overhead for small arrays in ptsort

jvanvugt:
`ptsort` is often called on small arrays, so I use a stackbuffer when the array is relatively small.
On my 6MP test image, this reduces the time taken in fit quads from 47.5ms to 44ms (averaged over 100 runs).

### DIFF
--- a/apriltag_quad_thresh.c
+++ b/apriltag_quad_thresh.c
@@ -714,8 +714,16 @@ static inline void ptsort(struct pt *pts, int sz)
 #undef MAYBE_SWAP
 
     // a merge sort with temp storage.
-
-    struct pt *tmp = malloc(sizeof(struct pt) * sz);
+    // Use stack allocation for small arrays to avoid malloc overhead
+    #define STACK_BUFFER_SIZE 256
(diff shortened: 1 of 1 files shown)


apriltag_quad_thresh.c
```diff
@@ -714,8 +714,16 @@ static inline void ptsort(struct pt *pts, int sz)
 #undef MAYBE_SWAP
 
     // a merge sort with temp storage.
-
-    struct pt *tmp = malloc(sizeof(struct pt) * sz);
+    // Use stack allocation for small arrays to avoid malloc overhead
+    #define STACK_BUFFER_SIZE 256
+    struct pt stack_buffer[STACK_BUFFER_SIZE];
+    struct pt *tmp;
+    bool use_heap = sz > STACK_BUFFER_SIZE;
```
christian-rauch, line 721: Make this a `const bool use_heap` as it is not changing afterwards. Squash this change into the same commit.

christian-rauch (changes requested):

christian-rauch (approved):

christian-rauch:
Thanks.

PR #404, opened 2025-10-07
Use path halving in unionfind to improve performance

jvanvugt:
Do path halving instead of path compression in unionfind. This makes the clustering take approximately 46ms instead of 52ms on my 6MP test image.


christian-rauch:
I converted to Draft, since you are still pushing to the PR. Once you are ready for the actual review, mark this PR as ready and let me know.

jvanvugt:
@christian-rauch I accidentally pushed to this branch instead of a new one for a different optimization. It's ready for your review now

PR #414, opened 2025-12-31
corrects capacity calculations

sidd-27:
This fixes the error in the capacity calculation math, where more than 6 times more space was being allocated than needed, because the capacity math used permutations instead of combination

### DIFF
--- a/apriltag.c
+++ b/apriltag.c
@@ -207,10 +207,10 @@ static void quick_decode_init(apriltag_family_t *family, int maxhamming)
         capacity += family->ncodes * nbits;
 
     if (maxhamming >= 2)
-        capacity += family->ncodes * nbits * (nbits-1);
+        capacity += family->ncodes * (nbits * (nbits-1)) / 2;
 
     if (maxhamming >= 3)
-        capacity += family->ncodes * nbits * (nbits-1) * (nbits-2);
+        capacity += family->ncodes * nbits * ((nbits-1) * (nbits-2)) / 6;
 
     qd->nentries = capacity * 3;
 


sidd-27:
I'm a maintainer at kornia-rs, which has its own native implementation of apriltag, and I made a few observations

The current structure for the table slots is -

```
struct quick_decode_entry
{
    uint64_t rcode;   // the queried code
    uint16_t id;      // the tag ID (a small integer)
    uint8_t hamming;  // how many errors corrected?
    uint8_t rotation; // number of rotations [0, 3]
};

```
which, ends up being 16 bytes long due to alignment, thus we lose 4 bytes to padding. To remedy this, we could decouple the rcode and the rest of the metadata to achieve something like this

struct QuickDecodeEntry {
    // [ ID (16) | Hamming (6) | Rot (2) ] stored in 3 bytes
    uint8_t data[3];
};

struct QuickDecode {
    vector<uint64_t > rcodes;
    vector<QuickDecodeEntry> entries;
};

storing exactly the amount of data we need in just 11 bytes per entry, resulting in a **31% improvement in memory usage**. Another benefit is more optimal cache utilisation, as during matching, we no longer need to load in the metadata, which pollutes the cache line, leading to faster lookups

i did a simple implementation of this in Rust, and these were the results I got -

Creation – Original: ~14.7 ms
Creation – Optimized: ~14.1 ms

Lookup – Original: ~872 µs
Lookup – Optimized: **~610 µs, ~30% faster**


Would you be interested in these optimizations?

christian-rauch:
> Additionally, another minor point: Is the rotation field in the quickdecode table effectively useless? It's never set anywhere and is immediately overwritten in the quick_decode_codeword function.

The `rotation` is set in `quick_decode_codeword` and used in `quad_decode_task`.

christian-rauch:
I can confirm that this roughly halves the memory requirements for the `tagStandard52h13` family.

PR #421, opened 2026-01-08
Reduced QuickDecode memory consumption by 128,884x

sidd-27:
### Problem Statement
The previous AprilTag decoding implementation relied on exhaustive precomputation of error permutations to achieve O(1) lookups. While effective for small tag families (e.g., 16h5), this approach is computationally and spatially prohibitive for larger families such as **52h13**.

For the 52h13 family, precomputing all variations with up to 2-bit errors results in approximately 67 million hash map entries. which amounts to approximately 3 GB of RAM being used for this simple check. If we were to extend this to 3-bit errors, the amount of space needed would increase exponentially to approximately 54 GB, making it essentially unusable. even for smaller tag families, dedicating this many resources to a single check is not possible in many small robotics applications

### Proposed Solution
This PR replaces the combinatorial precomputation strategy with a search-based algorithm utilizing the **Pigeonhole Principle**.

Instead of storing every possible error permutation, the new implementation indexes the valid tags by splitting the code into 4 discrete chunks (for eg 13 bits each for 52h13). Given a maximum tolerance of 3-bit errors, the Pigeonhole Principle guarantees that at least one of the four chunks in an observed tag must match the valid tag perfectly.

The decoding process is updated to:
1. Perform lookups on the 4 chunks to identify candidate tags.
2. Compute the Hamming distance between the observed code and the candidate's perfect code.
3. Return the match if the distance is within the specified threshold.

### Memory Complexity
The legacy implementation required storing every error permutation ($1 + N + \binom{N}{2} + \binom{N}{3}$) multiplied by a load factor of 3 to resolve collisions. For the 52h13 family, this necessitated allocating over **69000 slots per tag** (approx. 3 billion total slots). The new implementation eliminates this combinatorial explosion, storing exactly **4 references per tag**. **52h13 Memory Footprint went from ~54 GB to ~450 KB which is nearly a **128,884x** Reduction** 

### In summary - 
* **memory requirements and initialization:** Both of these are now several orders of magnitude better than the earlier implementation
* **Error correction:** now supports up to 3-bit error correction without any additional memory overhead; also, this technique can be scaled to test for 4 or even 5-bit errors with relative ease.
* **Regression:** Existing tests pass without modification.
* **performance:** as this is a technically more complex structure, the quickdecode function may be slightly slower than it used to be, but it still runs in nanoseconds and is not the bottleneck and there's no observable change to the overall performance

### DIFF
--- a/review.txt
+++ b/review.txt
@@ -1,5 +1,0 @@
-I think ENOMEM should already be set by calloc, so this is unnecessary. Please add a debug_print here to tell the user that memory allocation failed.
-
-This also applies to later instances of setting errno
-
-Can you make a const variable for NUM_CHUNKS instead of just putting 4
--- a/test/test_quick_decode.c
+++ b/test/test_quick_decode.c
@@ -41,14 +41,14 @@ void test_family(apriltag_family_t *fam) {
 
     int nbits = fam->nbits;
     
-    for (int i = 0; i < fam->ncodes; i++) {
+    for (uint32_t i = 0; i < fam->ncodes; i++) {
         uint64_t code = fam->codes[i];
         
         // Test 0 errors
         struct quick_decode_result res;
         quick_decode_codeword(fam, code, &res);
         if (res.id != i || res.hamming != 0) {
-            printf("Failed 0 errors: code %d, got id %d hamming %d\n", i, res.id, res.hamming);
+            printf("Failed 0 errors: code %u, got id %d hamming %d\n", i, res.id, res.hamming);
             exit(1);
         }
         
@@ -59,7 +59,7 @@ void test_family(apriltag_family_t *fam) {
             uint64_t c1 = code ^ (1ULL << b1);
             quick_decode_codeword(fam, c1, &res);
              if (res.id != i || res.hamming != 1) {
-                printf("Failed 1 error: code %d bit %d, got id %d hamming %d\n", i, b1, res.id, res.hamming);
+                printf("Failed 1 error: code %u bit %d, got id %d hamming %d\n", i, b1, res.id, res.hamming);
                 exit(1);
             }
             
@@ -70,7 +70,7 @@ void test_family(apriltag_family_t *fam) {
                 uint64_t c2 = c1 ^ (1ULL << b2);
                 quick_decode_codeword(fam, c2, &res);
                 if (res.id != i || res.hamming != 2) {
-                    printf("Failed 2 errors: code %d bits %d,%d, got id %d hamming %d\n", i, b1, b2, res.id, res.hamming);
+                    printf("Failed 2 errors: code %u bits %d,%d, got id %d hamming %d\n", i, b1, b2, res.id, res.hamming);
                     exit(1);
                 }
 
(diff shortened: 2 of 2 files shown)


christian-rauch:
Thank you a lot for this PR. Since this is quite involved, and we have to make a tradeoff between memory and computation speed for smaller tags, I am adding @mkrogius as a reviewer.

I also tried out the "Copilot" reviewer to see how useful this is. Some of the suggestions regarding the NULL pointer check indeed make sense to me.

christian-rauch (approved):

sidd-27:
I mean, for small code families, sure, a for loop iterating through all the codewords might just be the most efficient approach, but i suppose with larger families with 50 to 60k+ codewords, the lookup time increases, especially since we also need to look at 4 rotations for each codeword, and possibly multiple codewords per frame. This may still be pretty fast on modern machines, but it just won't scale as well, with larger and more complex tags imo

also, from what I've read, extending up to 5 or 6 bit error correction may just add false positives, but if needed, this code could be modified to be scalable like that with more chunks

sidd-27:
If further optimisation is needed, we could have it so that only the "canonical" forms, i.e., the ones with the lowest or highest numeric value, are stored in the lookup table. When we get an observed codeword, we just rotate it 4 times, find the canonical version, and look that up, instead of what we do rn, which is look up once per rotation, bringing the number of lookups needed from 2.5 on average to just one. And since rotate 90 is just bitwise operations, this would effectively make lookup 2.5x faster on average, I think.

apriltag.c
```diff
@@ -198,126 +203,130 @@ static void quick_decode_init(apriltag_family_t *family, int maxhamming)
...
-    for (uint32_t i = 0; i < family->ncodes; i++) {
-        uint64_t code = family->codes[i];
+    qd->shifts[0] = 0;
+    qd->shifts[1] = qd->chunk_size;
+    qd->shifts[2] = qd->chunk_size * 2;
+    qd->shifts[3] = qd->chunk_size * 3;
 
-        // add exact code (hamming = 0)
-        quick_decode_add(qd, code, i, 0);
+    for (int i = 0; i < 4; i++) {
```
mkrogius: Can you make a const variable for NUM_CHUNKS instead of just putting 4

apriltag.c
```diff
@@ -198,126 +203,130 @@ static void quick_decode_init(apriltag_family_t *family, int maxhamming)
...
+        qd->chunk_offsets[i] = calloc(qd->capacity + 1, sizeof(uint16_t));
```
mkrogius: Why capacity plus one?

mkrogius:
It would be nice to make the 52-bit tags more usable with 3-bit error correction. I will review this PR (In general I think we need much more complete test coverage for this project if we want to continue developing this repo).

@sidd-27 I do have a couple of high-level questions / comments:
* The implementation of quick_decode_init is different than I expected. I expected that we could simply create 4 hash tables, all using linear probing (just like the current hash table implementation). I think that your method is essentially the same but uses less memory.
* An alternative to chunking would be to use the current method, but instead of inserting all possible variations of all codes, insert all possible variations of the first `p` bits of each code. This would save some amount of memory and be a smaller change. Was this an option you considered?
* Would it be possible to add a test that all codes from all families and all errors of those codes (w/ up to 3 bits flipped) decode correctly?

apriltag.c
```diff
@@ -169,26 +163,37 @@ static struct quad *quad_copy(struct quad *quad)
...
-    qd->entries[bucket].hamming = hamming;
-}
+struct quick_decode
+{
+    int nbits;
+    int chunk_size;
+    int capacity;
+    int chunk_mask;
+    int shifts[4];
+    uint16_t* chunk_offsets[4];
```
mkrogius: Add comment: `chunk_offsets` is a map from chunk value to a range of locations in `chunk_ids`. Together with `chunk_ids`, this allows a lookup of all codes matching a chunk value.

PR #428, opened 2026-02-27
Add ArUco families support, deep-copy utilities, and pose ambiguity handling

fspindle:
This PR introduces three major improvements to the AprilTag library: native support for ArUco tag families, deep-copy utility functions for better memory management, and an API to retrieve the alternative pose solution during pose estimation.

### 1. Native ArUco Support

Integrated the ArUco dictionaries into the AprilTag ecosystem. This allows users to detect ArUco markers using the robust AprilTag 3 detector without needing external conversion tools.

- Added families: 4x4, 5x5, 6x6, 7x7 (various sizes from 50 to 1000) and ArUcoMIP36h12.
- Updated apriltag_pywrap, apriltag_demo, and opencv_demo to support these new families.
- Added comprehensive tests in test_quick_decode.c.

### 2. Deep-Copy Utilities

Added functions to simplify memory management when detections need to be passed between threads or stored beyond the detector's lifecycle.

- apriltag_detection_copy(): Deep copy of a single detection.
- apriltag_detections_copy(): Deep copy of a zarray_t of detections.
- apriltag_detector_copy(): Clones a detector configuration.

### 3. Pose Ambiguity Handling

Planar markers often have two mathematically plausible pose solutions. I've exposed the second solution from the orthogonal iteration algorithm.

- Added get_second_solution() to apriltag_pose.h.
- This allows developers to use temporal filtering or external constraints to choose the correct orientation when the error difference is small.

PR #429, opened 2026-03-05
Introduce deep-copy utilities

fspindle:
Added functions to simplify memory management when detections need to be passed between threads or stored beyond the detector's lifecycle.

    apriltag_detection_copy(): Deep copy of a single detection.
    apriltag_detections_copy(): Deep copy of a zarray_t of detections.
    apriltag_detector_copy(): Clones a detector configuration.

This PR follows recommendation given in PR  #428

### DIFF
--- a/apriltag.c
+++ b/apriltag.c
@@ -1496,3 +1496,58 @@ image_u8_t *apriltag_to_image(apriltag_family_t *fam, uint32_t idx)
     }
     return im;
 }
+
+void apriltag_detection_copy(apriltag_detection_t* src, apriltag_detection_t* dst)
+{
+    assert(src != NULL);
+    assert(dst != NULL);
+
+    if (dst->H) {
+        matd_destroy(dst->H);
+    }
+    dst->H = matd_copy(src->H);
+
+    dst->c[0] = src->c[0];
+    dst->c[1] = src->c[1];
+
+    for (int i = 0; i < 4; i++) {
+        for (int j = 0; j < 2; j++) {
+            dst->p[i][j] = src->p[i][j];
+        }
+    }
+
+    dst->id = src->id;
+    dst->family = src->family;
+    dst->hamming = src->hamming;
+    dst->decision_margin = src->decision_margin;
+}
+
+zarray_t* apriltag_detections_copy(zarray_t* detections)
+{
+    zarray_t* detections_copy = zarray_create(sizeof(apriltag_detection_t*));
+    for (int i = 0; i < zarray_size(detections); i++) {
+        apriltag_detection_t* det;
+        zarray_get(detections, i, &det);
+
+        apriltag_detection_t* det_copy = (apriltag_detection_t*)calloc(1, sizeof(apriltag_detection_t));
+        apriltag_detection_copy(det, det_copy);
+        zarray_add(detections_copy, &det_copy);
+    }
+
+    return detections_copy;
+}
+
+apriltag_detector_t *apriltag_detector_copy(apriltag_detector_t *src)
+{
+    apriltag_detector_t *dst = (apriltag_detector_t *)malloc(sizeof(apriltag_detector_t));
+    // Shallow copy of all scalar fields
+    *dst = *src;
+
+    // Reinitialize pointer fields to independent default values to avoid shared ownership and double-free issues
+    dst->tag_families = zarray_create(sizeof(apriltag_family_t *));
+    dst->tp = timeprofile_create();
+    dst->wp = workerpool_create(src->nthreads);
+
+    return dst;
+}
+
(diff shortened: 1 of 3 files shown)


apriltag.c
```diff
@@ -1496,3 +1496,66 @@ image_u8_t *apriltag_to_image(apriltag_family_t *fam, uint32_t idx)
...
+    td_copy->qtp.cos_critical_rad = td->qtp.cos_critical_rad;
+    td_copy->qtp.deglitch = td->qtp.deglitch;
+    td_copy->qtp.min_white_black_diff = td->qtp.min_white_black_diff;
+
+    td_copy->refine_edges = td->refine_edges;
+    td_copy->decode_sharpening = td->decode_sharpening;
+    td_copy->debug = td->debug;
+
+    return td_copy;
+}
```
christian-rauch: Since `apriltag_detector_t` is just a struct, why would someone use this function over just copying the struct?

Also, this does not copy the internal values, such as the `zarray_t *tag_families;`. Is this intended?

fspindle: > Since `apriltag_detector_t` is just a struct, why would someone use this function over just copying the struct?

Because `apriltag_detector_t` contains pointer fields (`tag_families`, `tp`) that are heap-allocated. A plain struct copy would only do a shallow copy, leaving both detectors sharing the same pointed-to data. When either one is destroyed, the other would hold dangling pointers. This function performs a proper deep copy by re-creating a fresh detector and only copying the scalar configuration fields, which is exactly the safe behavior needed for the copy constructor like the one implemented in ViSP [vpDetectorAprilTag](https://github.com/lagadic/visp/blob/master/modules/detection/src/tag/vpDetectorAprilTag.cpp) class.

> Also, this does not copy the internal values, such as the `zarray_t *tag_families;`. Is this intended?

This is intentional. `tag_families` holds pointers to `apriltag_family_t` objects that have their own lifecycle managed by `tagXXX_create()` / `tagXXX_destroy()`. Blindly copying the `zarray_t*` would result in two detectors sharing ownership of the same family objects, leading to a double-free on destruction. Instead, the copy constructor recreates the family from scratch using `tagXXX_create()` and can re-registers it via `apriltag_detector_add_family_bits()` as it is done in `vpApriltagDetector.cpp`, which achieves the same configuration while maintaining independent, safe ownership.

christian-rauch: You are right about the dangling pointers. A user would need to know which `apriltag_detector_t` is still using the `tag_families` array.

But at the moment, the `tag_families` is initialised with an empty array (via `apriltag_detector_create()`). If this is a deep copy, shouldn't the function also create a copy of the `tag_families` array? You mention double-free issues if two `apriltag_detector_t` instances share the same `tag_families` array. But this should not be the case with a "deep copy". I guess the same applies to `tp`.

If you do not use these fields, and thus you do not need a deep copy of `apriltag_detector_t` and just want to make sure that these pointers point to default initialised values and do not point to the original's `apriltag_detector_t` data, we could also set this to NULL and handle this case in the functions to make the intent more clear.

If you still want this to have the default values, I think I would still prefer a struct copy of `apriltag_detector_t` and then manually setting the default values via `zarray_create(sizeof(apriltag_family_t*))` and `timeprofile_create()`. This has the benefit, that the code is shorter and does not need to adapt if the `apriltag_detector_t` struct is extended with non-pointer fields.

fspindle: You are right. The current implementation is misleading since it is not a true deep copy, and it will silently break if new scalar fields are added to `apriltag_detector_t` in future versions.
Your last suggestion is the most robust approach: start with a struct copy to capture all scalar fields automatically, then explicitly reinitialize the pointer fields (`tag_families`, `tp` and `wp`) to fresh default values. This makes the intent clear and future-proof.

Here is what the updated implementation would look like (tested on my side):
```
apriltag_detector_t *apriltag_detector_copy(apriltag_detector_t *src)
{
  apriltag_detector_t *dst = (apriltag_detector_t *)malloc(sizeof(apriltag_detector_t));
  // Shallow copy of all scalar fields
  *dst = *src;

  // Reinitialize pointer fields to independent default values to avoid shared ownership and double-free issues
  dst->tag_families = zarray_create(sizeof(apriltag_family_t *));
  dst->tp = timeprofile_create();
  dst->wp = workerpool_create(src->nthreads);

  return dst;
}
```
Would this approach work for you?

christian-rauch: Yes. I think this is nicer than relisting all the struct fields separately. It also makes it a bit more clear, that this is not a "deep copy" but only copying the non-pointer elements. Can you squash this into your commit?

christian-rauch: I saw that you adjusted the function. When you are done with the proposed changes, just re-request a review so I see that you are done.

fspindle: Done

christian-rauch (approved):

PR #431, opened 2026-03-06
Add get_second_solution() to handle tag pose ambiguity

fspindle:
# Pose Ambiguity Handling

Planar markers often have two mathematically plausible pose solutions. This PR exposes the second solution from the orthogonal iteration algorithm.

- Added `get_second_solution()` to `apriltag_pose.h`. It allows to extract second pose candidate into a dedicated function. Uses `fix_pose_ambiguities()` to find the alternative solution and refines it via orthogonal iteration. Returns `HUGE_VAL` in `err2` if no second solution exists.
- This allows developers to use temporal filtering or external constraints to choose the correct orientation when the error difference is small.

### DIFF
--- a/README.md
+++ b/README.md
@@ -193,6 +193,21 @@ Note: The tag size should not be measured from the outside of the tag. The tag s
 ### Coordinate System
 The coordinate system has the origin at the camera center. The z-axis points from the camera center out the camera lens. The x-axis is to the right in the image taken by the camera, and y is down. The tag's coordinate frame is centered at the center of the tag. From the viewer's perspective, the x-axis is to the right, y-axis down, and z-axis is into the tag.
 
+### Handling Pose Ambiguity
+Planar targets often result in two possible pose solutions with similar errors (the "ambiguity" problem). To retrieve the alternative solution, you can use:
+
+```c
+apriltag_pose_t pose1, pose2;
+double err1 = estimate_tag_pose(&info, &pose1);
+double err2;
+int nIters = 50;
+
+// v and p are the object and image points used during the iteration
+get_second_solution(v, p, &pose1, &pose2, nIters, &err2);
+```
+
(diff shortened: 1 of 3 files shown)


christian-rauch:
@mkrogius Can you have a look at this?

mkrogius:
We already have `estimate_tag_pose_orthogonal_iteration` which returns both solutions. Although the name could be clearer, this seems more straightforward to call than `get_second_solution` since the user does not need to calculate the object and image points

christian-rauch:
@fspindle Can you reuse `estimate_tag_pose_orthogonal_iteration`:
https://github.com/AprilRobotics/apriltag/blob/305766652af34cafa5bab68fc2ebb2ca272e1482/apriltag_pose.c#L490-L523
instead of the custom functions proposed here?

fspindle:
@christian-rauch I will check end of the week. Thanks.

fspindle:
@christian-rauch @mkrogius I agree that the function I propose to introduce is not optimal. This PR can be closed.

I found that the ubuntu package doesn't expose  symbols of `estimate_pose_for_tag_homography()` and `estimate_tag_pose_orthogonal_iteration()`. Their visibility seems hidden. Could be nice to change that.

christian-rauch:
As far as I can see, we do not control the visibility of symbols.

Which Ubuntu package are you talking about? Looking at the [`apriltag` package manifest](https://salsa.debian.org/science-team/apriltag/-/blob/master/debian/libapriltag3t64.symbols?ref_type=heads) in Debian, they only expose `estimate_tag_pose@Base 3.1.7`. This is out of our control, and you have to open an issue there to get teh symbols added to the list.

christian-rauch:
@fspindle I am closing this at it appears that the functionality already exists.

PR #432, opened 2026-03-25
Micro optimizations (-15%)

bouk:
I set up Claude Code with an [autoresearch](https://github.com/karpathy/autoresearch)-style optimization loop to see what it could do for this image:

It ended up with a bunch of micro-optimizations that added up lead to a significant improvement. It also did a bunch of cheating by overfitting `apriltag_quad_thresh_params` to exactly this image, but I left that out of course 😄 

```
┌─────────┬───────────┬───────┬──────────────────────────────────────────────────────┐
│ commit  │ median_ms │ delta │                       message                        │
├─────────┼───────────┼───────┼──────────────────────────────────────────────────────┤
│ 3057666 │ 44.090    │ —     │ baseline (upstream/master)                           │
├─────────┼───────────┼───────┼──────────────────────────────────────────────────────┤
│ acfcdcd │ 41.199    │ -2.89 │ Pre-allocate cluster zarrays with capacity 64        │
├─────────┼───────────┼───────┼──────────────────────────────────────────────────────┤
│ 55c09b3 │ 40.798    │ -0.40 │ Optimize horizontal UF connect                       │
├─────────┼───────────┼───────┼──────────────────────────────────────────────────────┤
│ 24e3608 │ 39.703    │ -1.10 │ Remove sched_yield() from worker thread              │
├─────────┼───────────┼───────┼──────────────────────────────────────────────────────┤
│ 1c263eb │ 39.326    │ -0.38 │ Remove unnecessary mutex lock in workerpool_add_task │
├─────────┼───────────┼───────┼──────────────────────────────────────────────────────┤
│ 724ca6a │ 38.001    │ -1.33 │ Reduce the number of worker jobs                     │
├─────────┼───────────┼───────┼──────────────────────────────────────────────────────┤
│ 220a424 │ 37.614    │ -0.39 │ Fuse blur+threshold passes                           │
├─────────┼───────────┼───────┼──────────────────────────────────────────────────────┤
│ 2222ec2 │ 37.541    │ -0.07 │ Extract UF task params to local vars                 │
└─────────┴───────────┴───────┴──────────────────────────────────────────────────────┘
```

It ran `make -C build apriltag_demo -j8 && build/apriltag_demo -t 4 -i 5 -x 1.0 -f tagStandard52h13 img.jpg` where it had to make sure that the output was always the same.

2222ec2 seems the most questionable, perhaps we should drop that

### DIFF
--- a/apriltag_quad_thresh.c
+++ b/apriltag_quad_thresh.c
@@ -1635,6 +1635,7 @@ zarray_t* do_gradient_clusters(image_u8_t* threshim, int ts, int y0, int y1, int
                                                                             \
                             entry->id = clusterid;                          \
                             entry->cluster = zarray_create(sizeof(struct pt)); \
+                            zarray_ensure_capacity(entry->cluster, 64);     \
                             entry->next = clustermap[clustermap_bucket];    \
                             clustermap[clustermap_bucket] = entry;          \
                         }                                                   \
--- a/apriltag_quad_thresh.c
+++ b/apriltag_quad_thresh.c
@@ -108,31 +108,22 @@ struct cluster_task
 };
 
 struct minmax_task {
-    int ty;
+    int ty0, ty1;
 
     image_u8_t *im;
     uint8_t *im_max;
     uint8_t *im_min;
 };
 
-struct blur_task {
-    int ty;
-
-    image_u8_t *im;
-    uint8_t *im_max;
-    uint8_t *im_min;
-    uint8_t *im_max_tmp;
-    uint8_t *im_min_tmp;
-};
-
-struct threshold_task {
-    int ty;
+struct blur_threshold_task {
+    int ty0, ty1;
 
     apriltag_detector_t *td;
     image_u8_t *im;
     image_u8_t *threshim;
     uint8_t *im_max;
     uint8_t *im_min;
+    int tw, th;
 };
 
 struct remove_vertex
@@ -1104,10 +1095,10 @@ void do_minmax_task(void *p)
     const int tilesz = 4;
     struct minmax_task* task = (struct minmax_task*) p;
     int s = task->im->stride;
-    int ty = task->ty;
     int tw = task->im->width / tilesz;
     image_u8_t *im = task->im;
 
+    for (int ty = task->ty0; ty < task->ty1; ty++)
     for (int tx = 0; tx < tw; tx++) {
         uint8_t max = 0, min = 255;
 
@@ -1128,46 +1119,13 @@ void do_minmax_task(void *p)
     }
 }
 
-void do_blur_task(void *p)
+ 
+void do_blur_threshold_task(void *p)
 {
     const int tilesz = 4;
-    struct blur_task* task = (struct blur_task*) p;
-    int ty = task->ty;
-    int tw = task->im->width / tilesz;
-    int th = task->im->height / tilesz;
-    uint8_t *im_max = task->im_max;
-    uint8_t *im_min = task->im_min;
-
-    for (int tx = 0; tx < tw; tx++) {
-        uint8_t max = 0, min = 255;
-
-        for (int dy = -1; dy <= 1; dy++) {
-            if (ty+dy < 0 || ty+dy >= th)
-                continue;
-            for (int dx = -1; dx <= 1; dx++) {
-                if (tx+dx < 0 || tx+dx >= tw)
-                    continue;
-
-                uint8_t m = im_max[(ty+dy)*tw+tx+dx];
-                if (m > max)
-                    max = m;
-                m = im_min[(ty+dy)*tw+tx+dx];
-                if (m < min)
-                    min = m;
-            }
-        }
-
-        task->im_max_tmp[ty*tw + tx] = max;
-        task->im_min_tmp[ty*tw + tx] = min;
-    }
-}
-
-void do_threshold_task(void *p)
-{
-    const int tilesz = 4;
-    struct threshold_task* task = (struct threshold_task*) p;
-    int ty = task->ty;
-    int tw = task->im->width / tilesz;
+    struct blur_threshold_task* task = (struct blur_threshold_task*) p;
+    int tw = task->tw;
+    int th = task->th;
     int s = task->im->stride;
     uint8_t *im_max = task->im_max;
     uint8_t *im_min = task->im_min;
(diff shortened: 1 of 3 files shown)


apriltag_quad_thresh.c
```diff
@@ -1031,10 +1022,17 @@ static void do_unionfind_line2(unionfind_t *uf, image_u8_t *im, int w, int s, in
         if (v == 127)
             continue;
 
-        // (dx,dy) pairs for 8 connectivity:
-        // (-1, -1)    (0, -1)    (1, -1)
-        // (-1, 0)    (REFERENCE)
```
bouk, line 1036: I see a bunch of comments got lost, lmk if you want them back

christian-rauch: I guess we should keep the comments, if they are not part of removed code etc.

apriltag_quad_thresh.c
```diff
@@ -1031,10 +1031,17 @@ static void do_unionfind_line2(unionfind_t *uf, image_u8_t *im, int w, int s, in
...
+            uint32_t left_id = y*w + x - 1;
+            uint32_t cur_id = y*w + x;
+            uint32_t left_rep = unionfind_get_representative(uf, left_id);
+            if (uf->parent[cur_id] == 0xffffffff) {
+                uf->parent[cur_id] = left_rep;
+                uf->size[left_rep]++;
+            } else {
+                unionfind_connect(uf, cur_id, left_id);
+            }
+        }
```
christian-rauch: I don't understand what is happening here.

This replaces `DO_UNIONFIND2`, which does:
```
if (im->buf[(y + dy)*s + x + dx] == v)
  unionfind_connect(uf, y*w + x, (y + dy)*w + x + dx);
```

How is this optimised?

common/workerpool.c
```diff
@@ -81,7 +81,6 @@ void *worker_thread(void *p)
         zarray_get_volatile(wp->tasks, wp->taskspos, &task);
         wp->taskspos++;
         pthread_mutex_unlock(&wp->mutex);
-        sched_yield();
```
christian-rauch, line 84: Why is the "yield" not required anymore?

apriltag.c
```diff
@@ -1184,7 +1184,7 @@ zarray_t *apriltag_detector_detect(apriltag_detector_t *td, image_u8_t *im_orig)
     if (1) {
         image_u8_t *im_samples = td->debug ? image_u8_copy(im_orig) : NULL;
 
-        int chunksize = 1 + zarray_size(quads) / (APRILTAG_TASKS_PER_THREAD_TARGET * td->nthreads);
+        int chunksize = 1 + zarray_size(quads) / (1 * td->nthreads);
```
christian-rauch: Here, `APRILTAG_TASKS_PER_THREAD_TARGET` is replaced by 1 ...

apriltag_quad_thresh.c
```diff
@@ -1840,7 +1856,7 @@ zarray_t* fit_quads(apriltag_detector_t *td, int w, int h, zarray_t* clusters, i
     }
 
     int sz = zarray_size(clusters);
-    int chunksize = 1 + sz / (APRILTAG_TASKS_PER_THREAD_TARGET * td->nthreads);
+    int chunksize = 1 + sz / (4 * td->nthreads);
```
christian-rauch: ... and here `APRILTAG_TASKS_PER_THREAD_TARGET` is replaced by 4.

apriltag_quad_thresh.c
```diff
@@ -115,18 +115,7 @@ struct minmax_task {
...
-struct blur_task {
-    int ty0, ty1;
-
-    image_u8_t *im;
-    uint8_t *im_max;
-    uint8_t *im_min;
-    uint8_t *im_max_tmp;
-    uint8_t *im_min_tmp;
-    int tw, th;
-};
```
christian-rauch, line 126: This was changed in the previous commit. If you touch the code multiple times, merge the commits.

apriltag_quad_thresh.c
```diff
@@ -1054,8 +1054,11 @@ static void do_unionfind_task2(void *p)
 {
     struct unionfind_task *task = (struct unionfind_task*) p;
 
+    unionfind_t *uf = task->uf;
```
christian-rauch: The commit message says "Extract UF task params to local vars". What is "UF"? And with "I think this allows it to be inlined and unrolled" it's not really clear if there is a benefit to this change.

christian-rauch (changes requested):
Hm. Optimisations are always good. And I am not against AI usage to help with coding per se.

But do you actually understand, how the code is optimised? Some things like pre-allocation or removing threading locks are reasonable. But the changes in some commits are difficult to understand.

Specifically the "Reduce the number of worker jobs" commit is questionable. If this statement would be true, then it should suffice to change the `APRILTAG_TASKS_PER_THREAD_TARGET` from the default 10 to some more optimal value.

Also a commit message that starts with "I think" is not really trustworthy.

I would suggest that you extract the commits, where you can explain the changes in your own words in a commit message, and propose those for merging first.

bouk:
Yes, I will reword the commits, just wanted to make sure you're open to these changes.

bouk:
> Specifically the "Reduce the number of worker jobs" commit is questionable. If this statement would be true, then it should suffice to change the APRILTAG_TASKS_PER_THREAD_TARGET from the default 10 to some more optimal value.

True, but in my testing (I also did a bunch of manual testing) I found that there's different optimal values for the different tasks. But perhaps this is too granular and we should just pick 2 or 4 and make sure it's used everywhere

common/workerpool.c
bouk: I'm not sure why it's needed at all. Reading the man page: https://man7.org/linux/man-pages/man2/sched_yield.2.html

> Avoid calling sched_yield() unnecessarily or inappropriately
>       (e.g., when resources needed by other schedulable threads are
>        still held by the caller), since doing so will result in
>        unnecessary context switches, which will degrade system
>        performance.

It does have a significant performance impact

apriltag.c
bouk: Yes. It should just be `td->nthreads`

apriltag_quad_thresh.c
bouk: UF = 'unionfind'

I will drop this commit

apriltag_quad_thresh.c
bouk: if you're fine with both changes, sure. I'll squash them

apriltag_quad_thresh.c
bouk: Sorry, you're totally right...

I'll drop this change

bouk:
I'll make separate PRs for the changes, thanks for the feedback

bouk:
I made https://github.com/AprilRobotics/apriltag/pull/435, after better benchmarking this seems to be the only actual speedup that was in this PR

PR #433, opened 2026-03-26
Use min_cluster_pixels for early gradient cluster rejection

bouk:
Replace the hardcoded threshold of 25 pixels in do_gradient_clusters with the configurable min_cluster_pixels parameter. This rejects clusters earlier, reducing work later.

Is 1.18x faster in a bench with a 3088x2064 image with min_cluster_pixels=100

### DIFF
--- a/apriltag_quad_thresh.c
+++ b/apriltag_quad_thresh.c
@@ -102,6 +102,7 @@ struct cluster_task
     int w;
     int s;
     int nclustermap;
+    int min_cluster_pixels;
     unionfind_t* uf;
     image_u8_t* im;
     zarray_t* clusters;
@@ -1557,7 +1558,7 @@ unionfind_t* connected_components(apriltag_detector_t *td, image_u8_t* threshim,
     return uf;
 }
 
-zarray_t* do_gradient_clusters(image_u8_t* threshim, int ts, int y0, int y1, int w, int nclustermap, unionfind_t* uf, zarray_t* clusters) {
+zarray_t* do_gradient_clusters(image_u8_t* threshim, int ts, int y0, int y1, int w, int nclustermap, int min_cluster_pixels, unionfind_t* uf, zarray_t* clusters) {
     struct uint64_zarray_entry **clustermap = calloc(nclustermap, sizeof(struct uint64_zarray_entry*));
 
     int mem_chunk_size = 2048;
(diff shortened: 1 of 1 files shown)


apriltag_quad_thresh.c
```diff
@@ -1578,7 +1579,7 @@ zarray_t* do_gradient_clusters(image_u8_t* threshim, int ts, int y0, int y1, int
 
             // XXX don't query this until we know we need it?
             uint64_t rep0 = unionfind_get_representative(uf, y*w + x);
-            if (unionfind_get_set_size(uf, rep0) < 25) {
+            if ((int)unionfind_get_set_size(uf, rep0) < min_cluster_pixels) {
```
bouk, line 1582: Should this be `< min_cluster_pixels - 1`? To match the old behavior when `min_cluster_pixels = 24` (the default)

apriltag_quad_thresh.c
```diff
@@ -1610,7 +1611,7 @@ zarray_t* do_gradient_clusters(image_u8_t* threshim, int ts, int y0, int y1, int
                                                                         \
                 if (v0 + v1 == 255) {                                   \
                     uint64_t rep1 = unionfind_get_representative(uf, (y + dy)*w + x + dx); \
-                    if (unionfind_get_set_size(uf, rep1) > 24) {        \
+                    if ((int)unionfind_get_set_size(uf, rep1) >= min_cluster_pixels) { \
```
bouk, line 1614: same as above, should this be `> min_cluster_pixels`? This version makes more sense to me

fallenmi (changes requested):
The new early filter changes the meaning of `min_cluster_pixels` and can discard a quad that passes the existing final cluster-size check.

I built exact base `305766652af34cafa5bab68fc2ebb2ca272e1482` and exact head `33851747f36805d7b23554117a94e4c258383382` in Release mode. Both complete the repository's seven CTest cases. I then used the public detector API on the repository image `test/data/34139872896_defdb2f8d9_c.jpg`, with `tag36h11`, `quad_decimate = 1`, and `qtp.min_cluster_pixels = 250`:

```text
                         nthreads=1    nthreads=4
exact base detections         10            10
exact head detections          9             9
```

The lost base detection is stable across ten repeated single-thread runs and is centered at approximately `(680.597, 365.306)`. The current code applies `min_cluster_pixels` to the boundary-point cluster passed to `fit_quads`; this patch also requires each adjacent union-find component's area to meet the same value. Those are different quantities, so the component-area prefilter is not a safe proxy for the later quad-cluster threshold.

Please retain a semantics-preserving lower bound for the early component filter (or derive/prove a safe relationship) and add this non-default-parameter case as a regression test.

Reviewed by OpenAI Codex for @fallenmi at the exact head above. The oracle was external; no repository files were changed.

PR #434, opened 2026-03-26
Cache unionfind structure across detect calls and lazy-initialize size array

bouk:
For large images (3088x2064) this gives me an 1.33x(!) speedup.

Surprisingly only on Linux, on my macbook there's no difference

### DIFF
--- a/apriltag_quad_thresh.c
+++ b/apriltag_quad_thresh.c
@@ -1511,7 +1511,17 @@ image_u8_t *threshold_bayer(apriltag_detector_t *td, image_u8_t *im)
 }
 
 unionfind_t* connected_components(apriltag_detector_t *td, image_u8_t* threshim, int w, int h, int ts) {
-    unionfind_t *uf = unionfind_create(w * h);
+    uint32_t maxid = w * h;
+    if (td->cached_uf) {
+        if (td->cached_uf->maxid >= maxid) {
+            unionfind_reset(td->cached_uf);
+        } else {
+            unionfind_resize(td->cached_uf, maxid);
+        }
+    } else {
+        td->cached_uf = unionfind_create(maxid);
+    }
+    unionfind_t *uf = td->cached_uf;
 
     if (td->nthreads <= 1) {
         do_unionfind_first_line(uf, threshim, w, ts);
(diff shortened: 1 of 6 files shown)


Gold856:
apriltag_demo segfaults if I pass in two images on the second image, with the first one having a smaller resolution (640x400, followed by 1280x951). Switching the order such that the first image has the larger resolution does not trigger a segfault.

bouk:
@Gold856 you're right, I inverted a condition. Fixed now and I added a regression test

apriltag_quad_thresh.c
```diff
@@ -1513,7 +1513,7 @@ image_u8_t *threshold_bayer(apriltag_detector_t *td, image_u8_t *im)
 unionfind_t* connected_components(apriltag_detector_t *td, image_u8_t* threshim, int w, int h, int ts) {
     uint32_t maxid = w * h;
     if (td->cached_uf) {
-        if (td->cached_uf->maxid <= maxid) {
+        if (td->cached_uf->maxid >= maxid) {
```
christian-rauch, line 1516: Squash fixup commits into the original commits.

christian-rauch:
Can you rebase this and provide some details in the commit message what this change does and how it works?

bouk:
@christian-rauch I've rebased and added a little bit more detail in the commit

christian-rauch:
@mkrogius

common/unionfind.h
```diff
@@ -67,15 +67,27 @@ static inline void uf_store_parent(unionfind_t *uf, uint32_t id, uint32_t val)
...
-        uf->data[i].size = 0;
-    }
+    unionfind_reset(uf);
+    return uf;
+}
+
+static inline unionfind_t *unionfind_resize(unionfind_t *uf, uint32_t maxid)
+{
+    uf->maxid = maxid;
+    uf->data = (struct unionfind_node *) realloc(uf->data, (maxid+1) * sizeof(struct unionfind_node));
```
mkrogius: Why not free + malloc here? realloc will copy data unnecessarily

bouk: Good point, fixed

PR #439, opened 2026-04-20
Fix data race in unionfind_get_representative using relaxed atomics

joshi-bharat:
Multiple worker threads call unionfind_get_representative concurrently during do_gradient_clusters. The path-halving writes race with reads from other threads, triggering ThreadSanitizer warnings. Use __atomic_load_n/__atomic_store_n with __ATOMIC_RELAXED to make these accesses TSan-clean. On x86, relaxed atomics compile to plain mov instructions with zero performance cost.

Also adds a Google Benchmark target (BUILD_BENCHMARK=ON) using test/data images for performance regression testing.

### DIFF
--- a/common/unionfind.h
+++ b/common/unionfind.h
@@ -49,6 +49,24 @@ struct unionfind
     struct unionfind_node *data;
 };
 
+static inline uint32_t uf_load_parent(unionfind_t *uf, uint32_t id)
+{
+#if defined(__GNUC__) || defined(__clang__)
+    return __atomic_load_n(&uf->data[id].parent, __ATOMIC_RELAXED);
+#else
+    return uf->data[id].parent;
+#endif
+}
+
+static inline void uf_store_parent(unionfind_t *uf, uint32_t id, uint32_t val)
+{
+#if defined(__GNUC__) || defined(__clang__)
+    __atomic_store_n(&uf->data[id].parent, val, __ATOMIC_RELAXED);
+#else
+    uf->data[id].parent = val;
+#endif
+}
+
 static inline unionfind_t *unionfind_create(uint32_t maxid)
 {
     unionfind_t *uf = (unionfind_t*) calloc(1, sizeof(unionfind_t));
@@ -89,15 +107,15 @@ static inline uint32_t unionfind_get_representative(unionfind_t *uf, uint32_t id
 static inline uint32_t unionfind_get_representative(unionfind_t *uf, uint32_t id)
 {
     // unititialized node, so set to self
-    if (uf->data[id].parent == 0xffffffff) {
-        uf->data[id].parent = id;
+    if (uf_load_parent(uf, id) == 0xffffffff) {
+        uf_store_parent(uf, id, id);
         return id;
     }
 
     // Path halving: make every node point to its grandparent (single pass)
-    while (uf->data[id].parent != id) {
-        uf->data[id].parent = uf->data[uf->data[id].parent].parent;
-        id = uf->data[id].parent;
+    while (uf_load_parent(uf, id) != id) {
+        uf_store_parent(uf, id, uf_load_parent(uf, uf_load_parent(uf, id)));
+        id = uf_load_parent(uf, id);
     }
 
     return id;


christian-rauch (changes requested):
Thanks for the fix.

Since the benchmark is OFF by default and does not run in the CI, it will be hard to notice when this breaks. In any case, I prefer to have a PR with the fix, separately from a new feature. Or at least use separate commits for fixes and features.

Also, I would only add the benchmark, if this compiles and runs in the CI by default to avoid unnoticed breakage.

christian-rauch:
Please squash your "fixup" commit into the first one.

common/unionfind.h
```diff
@@ -89,15 +107,17 @@ static inline uint32_t unionfind_get_representative(unionfind_t *uf, uint32_t id
 static inline uint32_t unionfind_get_representative(unionfind_t *uf, uint32_t id)
 {
     // unititialized node, so set to self
-    if (uf->data[id].parent == 0xffffffff) {
-        uf->data[id].parent = id;
+    uint32_t parent = uf_load_parent(uf, id);
+    if (parent == 0xffffffff) {
```
christian-rauch: Do you have to store the `parent` here? In the while loop below, you use `uf_load_parent` directly.

common/unionfind.h
```diff
@@ -89,15 +107,17 @@ static inline uint32_t unionfind_get_representative(unionfind_t *uf, uint32_t id
...
         return id;
     }
 
     // Path halving: make every node point to its grandparent (single pass)
-    while (uf->data[id].parent != id) {
-        uf->data[id].parent = uf->data[uf->data[id].parent].parent;
-        id = uf->data[id].parent;
+    while (uf_load_parent(uf, id) != id) {
+        uint32_t grandparent = uf_load_parent(uf, uf_load_parent(uf, id));
+        uf_store_parent(uf, id, grandparent);
```
christian-rauch: Also here, do you have to store `grandparent`? It looks to me as if you can just nest the function calls. Is this for better readability?

common/unionfind.h
joshi-bharat: I don't think we need to save parent. I will remove this.

common/unionfind.h
joshi-bharat: this is just for readability. We can directly use it inside the store parent.

common/unionfind.h
christian-rauch: Here and with my comment below, I would just not store things that are reused directly and only once. Just chain them.

christian-rauch (approved):
Thanks a lot. This looks good now. I will merge this once the CI passes.

joshi-bharat:
FYI: I ran the benchmark a bunch of times. I did not find any statistically significant change in computation time. Actual runtime might vary based on system load.

# Benchmark Comparison: TSan-safe atomics fix

**Config:** 1000 iterations, -O3 Release, Google Benchmark, 24x CPU @ 5100 MHz

| Image | Threads | Master (ms) | With fix (ms) | Delta |
|-------|---------|-------------|---------------|-------|
| 33369213973_9d9bb4cc96_c.jpg (9 det) | 1 | 3.09 | 3.27 | +5.8% |
| 33369213973_9d9bb4cc96_c.jpg (9 det) | 2 | 1.96 | 2.06 | +5.1% |
| 33369213973_9d9bb4cc96_c.jpg (9 det) | 4 | 1.39 | 1.47 | +5.8% |
| 34085369442_304b6bafd9_c.jpg (18 det) | 1 | 5.70 | 5.89 | +3.3% |
| 34085369442_304b6bafd9_c.jpg (18 det) | 2 | 3.40 | 3.49 | +2.6% |
| 34085369442_304b6bafd9_c.jpg (18 det) | 4 | 2.16 | 2.19 | +1.4% |
| 34139872896_defdb2f8d9_c.jpg (7 det) | 1 | 3.38 | 3.74 | +10.7% |
| 34139872896_defdb2f8d9_c.jpg (7 det) | 2 | 2.04 | 2.14 | +4.9% |
| 34139872896_defdb2f8d9_c.jpg (7 det) | 4 | 1.43 | 1.49 | +4.2% |

**Note:** `__ATOMIC_RELAXED` on x86 compiles to identical `mov` instructions as plain loads/stores. Deltas are within run-to-run variance (CPU scaling was enabled). No consistent performance regression.

joshi-bharat:
[benchmark_master.csv](https://github.com/user-attachments/files/26943305/benchmark_master.csv)
[benchmark_with_atomic_fix.csv](https://github.com/user-attachments/files/26943306/benchmark_with_atomic_fix.csv)

christian-rauch:
> I did not find any statistically significant change in computation time. Actual runtime might vary based on system load.

Well, I think 10% increase might be significant for some.

I am actually surprised to see that this is consistently slower, given that it should just compile to a mov instruction. Also, I am wondering why the increase is the largest for smaller number of threads and fewer detections. Shouldn't blocking occur more often the more detections and threads are involved?

Do you have an idea why the processing time increases with fewer threads and detections?

joshi-bharat:
I am saying this is not significant because I ran the same benchmark like an hour ago on the same machine and it was showing that fix increased the speed even in 1000 iterations. So I am not sure we can trust this unless we ran on a machine that we can pin other computational processes.

